PR #8, opened 2019-10-18
Optimizing build.gradle

f-lombardo:
Please check if this gradle configuration works in IDEA

### DIFF
--- a/rpgJavaInterpreter-core/build.gradle
+++ b/rpgJavaInterpreter-core/build.gradle
@@ -59,12 +59,12 @@ generateGrammarSource {
     arguments += ['-package', 'com.smeup.rpgparser']
     outputDirectory = file(generatedMain + "/com/smeup/rpgparser")
 }
-compileKotlin.dependsOn generateGrammarSource
 compileJava.dependsOn generateGrammarSource
 compileJava.source generatedMainFile, sourceSets.main.java
-compileKotlin.source generatedMainFile, sourceSets.main.java, sourceSets.main.kotlin
 compileKotlin {
+    source generatedMainFile, sourceSets.main.java, sourceSets.main.kotlin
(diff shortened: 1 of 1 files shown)


rpgJavaInterpreter-core/build.gradle
```diff
@@ -59,12 +59,12 @@ generateGrammarSource {
     arguments += ['-package', 'com.smeup.rpgparser']
     outputDirectory = file(generatedMain + "/com/smeup/rpgparser")
 }
-compileKotlin.dependsOn generateGrammarSource
 compileJava.dependsOn generateGrammarSource
```
ftomassetti: Probably we can remove this and the following line

rpgJavaInterpreter-core/build.gradle
```diff
@@ -59,12 +59,12 @@ generateGrammarSource {
...
 compileJava.source generatedMainFile, sourceSets.main.java
-compileKotlin.source generatedMainFile, sourceSets.main.java, sourceSets.main.kotlin
 compileKotlin {
+    source generatedMainFile, sourceSets.main.java, sourceSets.main.kotlin
```
ftomassetti: good idea. However I am not sure if we need `sourceSets.main.java` as we are not using any Java code, I think

rpgJavaInterpreter-core/build.gradle
```diff
@@ -59,12 +59,12 @@ generateGrammarSource {
...
     kotlinOptions.freeCompilerArgs += ["-Xuse-experimental=kotlin.ExperimentalUnsignedTypes"]
+    dependsOn generateGrammarSource
```
ftomassetti: It makes sense. Personally I prefer to have the dependsOn clauses outside the single tasks because to me it makes more evident the relations between tasks, but it is a matter of test and this make the script cleaner imho

ftomassetti:
Probably you want to reopen it and press merge, not close

f-lombardo:
OK, thank you

PR #9, opened 2019-10-21
Fl01

f-lombardo:
Implementing methods for DB access


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
```diff
@@ -31,29 +32,45 @@ class DBSQLFile(private val name: String, private val connection: Connection) :
...
-        if (resultSet == null) {
-            chain(emptyList())
+        return readEqual(toFields(key))
+    }
+
+    private fun filterRecord(result: Record): Record {
+        return if (result.matches(lastKey)) {
+            result
+        } else {
+            resultSet?.last()
```
ftomassetti: Is this acceptable to have a null resultSet here?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
```diff
@@ -4,12 +4,13 @@ import com.smeup.rpgparser.interpreter.DBFile
 import com.smeup.rpgparser.interpreter.Field
 import com.smeup.rpgparser.interpreter.Record
 import com.smeup.rpgparser.interpreter.Value
-import java.lang.RuntimeException
 import java.sql.Connection
 import java.sql.ResultSet
 
 class DBSQLFile(private val name: String, private val connection: Connection) : DBFile {
```
ftomassetti: Given I am not familiar with this part of the code I looked into DBFile. Depending on the background of other contributors what DBFile does could be not obvious and we may want to add a few comments

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
```diff
@@ -4,12 +4,13 @@ import com.smeup.rpgparser.interpreter.DBFile
...
     private var resultSet: ResultSet? = null
+    private var lastKey: List<Field> = emptyList()
```
ftomassetti: is this specific for the readEqual? Or will this be updated by multiple operations?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
```diff
@@ -19,7 +20,7 @@ class DBSQLFile(private val name: String, private val connection: Connection) :
         if (resultSet == null) {
```
ftomassetti: You could use require here. I personally find it more obvious

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
```diff
@@ -31,29 +32,45 @@ class DBSQLFile(private val name: String, private val connection: Connection) :
...
+        return filterRecord(result)
     }
 
     override fun eof(): Boolean = resultSet?.isLast ?: false
 
     override fun chain(key: Value): Record {
+        return chain(toFields(key))
+    }
+
+    private fun toFields(keyValue: Value): List<Field> {
```
ftomassetti: It is not clear to me what this function does. Probably because I am not familiar with this portion of the codebase

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
```diff
@@ -31,29 +32,45 @@ class DBSQLFile(private val name: String, private val connection: Connection) :
...
+            readFromPositionedResultSet()
+        }
+        lastKey = keys
+        return filterRecord(result)
     }
 
     override fun eof(): Boolean = resultSet?.isLast ?: false
 
     override fun chain(key: Value): Record {
+        return chain(toFields(key))
```
ftomassetti: if the readEqual and chain methods taking a Value always delegate to the ones taking a list of fields we could move them into abstract classes or implement them directly into `DBFile` as they would not be specific to this class

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
```diff
@@ -31,29 +32,45 @@ class DBSQLFile(private val name: String, private val connection: Connection) :
...
+
+    private fun toFields(keyValue: Value): List<Field> {
         val keyName = keys.first()
-        return chain(listOf(Field(keyName, key)))
+        return listOf(Field(keyName, keyValue))
     }
 
     override fun chain(keys: List<Field>): Record {
         val keyNames = keys.map { it.name }
         val sql = "SELECT * FROM $name ${keyNames.whereSQL()} ${keyNames.orderBySQL()}"
```
ftomassetti: This looks good to me as it is, but if the complexity of the SQL statements increase we could consider using a library or creating a simple internal DSL to build statements

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/system_interface.kt
```diff
@@ -39,7 +41,20 @@ interface DBInterface {
...
 data class Field(val name: String, val value: Value)
 
-class Record(vararg fields: Field) : ArrayList<Field>(fields.asList())
+class Record(vararg fields: Field) : LinkedHashMap<String, Value>() {
+    init {
+        fields.forEach {
+            add(it)
+        }
+    }
+    fun matches(keyFields: List<Field>): Boolean {
```
ftomassetti: I would add a newline before matches

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/system_interface.kt
```diff
@@ -39,7 +41,20 @@ interface DBInterface {
...
+        return keyFields.all { this[it.name] == it.value }
```
ftomassetti: given the body is an expression you could rewrite it using the shortcut with the equals:

```
fun matches(keyFields: List<Field>) = keyFields.all { this[it.name] == it.value }
```

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/db/sql/DBSQLTest.kt
```diff
@@ -48,7 +48,8 @@ class DBSQLTest {
         assertTrue(dbFile!!.chain(StringValue("ABC")).isEmpty())
         val chainedRecord = dbFile.chain(StringValue("XXX"))
         assertEquals(2, chainedRecord.size)
-        assertEquals(StringValue("XXX"), chainedRecord[0].value)
-        assertEquals(DecimalValue(bigDecimalValue), chainedRecord[1].value)
+        val fieldsIterator = chainedRecord.iterator()
+        assertEquals(StringValue("XXX"), fieldsIterator.next().value)
+        assertEquals(DecimalValue(bigDecimalValue), fieldsIterator.next().value)
     }
 }
```
ftomassetti: Normally I would add a newline at the end of the file

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/interpreter/RecordTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import kotlin.test.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class RecordTest {
+
+    @Test
+    fun matchesOK() {
+        val record = Record(
+            Field("F1", StringValue("A")),
```
ftomassetti: I wonder if you could use a Pair instead of Field

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/interpreter/RecordTest.kt
```diff
@@ -0,0 +1,36 @@
...
+            Field("F2", IntValue(8)),
+            Field("F3", StringValue("W"))
+        )
+        val keyFieldsKO = listOf(
+            Field("F2", IntValue(8)),
+            Field("F3", StringValue("XXXXXXX"))
+        )
+        assertFalse(record.matches((keyFieldsKO)))
+    }
+}
```
ftomassetti: I would add a newline here

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
f-lombardo: At the moment filterRecord is called just when resultSet is not null, but, as I think it could be used elsewhere, I think we can have a "relaxed" behaviour. The idea is "in case you have an open resultset, just set on EOF, if you don't have it, EOF is already on". Do you think it wolud be better to encapsulate this in a fun with a speaking name?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
f-lombardo: Good question. The answer is "Since I'm learning RPG while creating the interpreter, I don't really know" :-) 
Should we write the code differently depending on this?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
f-lombardo: You're right!

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
ftomassetti: I am not sure as I do not know the context very well, but if you would be "surprised" if resultSet is actually null then I would write `resultSet!!.last()` and if later you confirm it can be relaxed then I would change it back

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
ftomassetti: no, just maybe a more specific name like `readEqualLastKey`

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
f-lombardo: A chain operation (lookup a record in a table) can be perfomed using one or more keys. So we have to transform a value (e.g. "ABC") in a Field (i.e. "PART_NUMBER"->"ABC") and then encapsulate  it in a list. Do you have any suggestion in order to make the code more clear?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/system_interface.kt
f-lombardo: OK

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/system_interface.kt
f-lombardo: Good idea

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/interpreter/RecordTest.kt
f-lombardo: Originally I wrote the code using Pair, but then the it seemed a bit obscure to me: I think that 
x.name, x.value are more understandable than x.first, x.second, see for example the matches fun

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
f-lombardo: We can't put it in DBFile, since this interface could (I hope) have different implementation in the future, for example with a NOSQL DB. 
IMHO an abstract class coluld be overkill here

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
f-lombardo: Good point. I started using a simple library for creating SQL statements, but it was too heavy at the moment. Let's see how the code evolves.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/db/sql/DBSQLFile.kt
ftomassetti: To me is not clear what a Field is in this context. Is it a data structure field? In your example is "PART_NUMBER" the name of the field used as a key and is "ABC" the actual value of the field?

ftomassetti: Also, the term Field makes me think to the data structure fields, while these are different fields, right? Perhaps we could have a more specific name to make this more obvious

PR #11, opened 2019-10-23
Mute fail

f-lombardo:
Introduced a FAIL MUTE annotation. At the moment it requires square brackets to work, but I'd like to use standard ones, as for the other annotations. To be honest, I don't really understand why the assertion 
MU* VAL1(A) VAL2(5) COMP(EQ)
could work, as in the MuteLexer there is this definition:
EXP: '['.*?']';


docs/mute.md
```diff
@@ -43,7 +43,20 @@ Note that the assertion is verified at the end of the program, so it doesn't sto
 You can add many ```TIMEOUT``` assertion in the same program, but, obviously, just the one with the shortest timeout is considered.
 
 There isn't a mandatory position to insert the ```TIMEOUT``` annotation line, that, as other ```MUTE``` annotations, are just considered as comments by the standard interpreter engine.
- 
+
+### Failing assertions 
+Like in other frameworks for unit testing, there is an assertion that always fails:
```
ftomassetti: We could add a newline between the title and the paragrah

docs/mute.md
```diff
@@ -43,7 +43,20 @@ Note that the assertion is verified at the end of the program, so it doesn't sto
...
+
+### Failing assertions 
+Like in other frameworks for unit testing, there is an assertion that always fails: 
+
+```
+    MU* FAIL('Something went wrong')
+```
+
+An expression can be used for the failing message:
+```
```
ftomassetti: I thing we can add a new line also begin the code section

docs/mute.md
```diff
@@ -43,7 +43,20 @@ Note that the assertion is verified at the end of the program, so it doesn't sto
...
+    MU* FAIL('Something went wrong')
+```
+
+An expression can be used for the failing message:
+``` 
+      D MSG             S             15    INZ('Failure message')
+     MU* FAIL(MSG)
+ ```
+
```
ftomassetti: We could add an example to explain a case in which is useful. For example to indicate that a certain portion of the code is not expected to be executed.

Also, is there any limitation? Can the FAIL annotation be used anywhere or does it need to be put in front of a statement?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/mute.kt
```diff
@@ -32,4 +39,14 @@ data class MuteAnnotationResolved(val muteLine: Int, val statementLine: Int)
  * The result of executing a mute annotation. Note that currently we have only annotations with two values.
  * This could change in the future.
  */
-data class MuteAnnotationExecuted(val programName: String, val expression: Expression, val value1Expression: Expression, val value2Expression: Expression, val result: Value, val value1Result: Value, val value2Result: Value)
+data class MuteAnnotationExecuted(val programName: String, val expression: Expression, val value1Expression: Expression, val value2Expression: Expression, val result: BooleanValue, val value1Result: Value, val value2Result: Value) {
+    fun succeeded(): Boolean = result.value
+
+    companion object {
+        private val DUMMYEXPRESSION = StringLiteral("")
```
ftomassetti: these fields seem suspicious. Perhaps we could make the parameters in `MuteAnnotationExecuted` nullable instead, or change this class to be abstract and have multiple subclasses

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/MuteExecutionTest.kt
```diff
@@ -75,638 +70,26 @@ class MuteExecutionTest {
...
-        val si = CollectorSystemInterface()
-        val rpgProgram = RpgProgram(cu)
-        rpgProgram.execute(si, linkedMapOf("ppdat" to StringValue("10")))
-        assertEquals(1, rpgProgram.params().size)
-        assertEquals(ProgramParam("ppdat", StringType(8)), rpgProgram.params()[0])
-        assertEquals(listOf("FIBONACCI OF: 10 IS: 55"), si.displayed)
-    }
-
-    @Test
-    fun executeHELLOCASE() {
```
ftomassetti: Why are we removing all these tests?

rpgJavaInterpreter-core/src/test/resources/mute/SIMPLE_MUTE_FAIL_EVALUATED_MESSAGE.rpgle
```diff
@@ -0,0 +1,8 @@
+     DA                S              8  0 INZ(5)
+     DB                S              8  0 INZ(8)
+     D RESULT          S              8  0 INZ(0)
+     D MSG             S             15    INZ('Failure message')
+      *
+    MU* FAIL[MSG]
+     C                   EVAL      RESULT = A + B
+     C                   SETON                                        LR
```
ftomassetti: I would add a newline at the end of the file

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/mute.kt
```diff
@@ -23,17 +23,21 @@ data class MuteAnnotationExecutionLogEntry(override val programName: String, val
...
-                    .expression()
-                    .toAst(conf)
+    fun extractExpressionFrom(token: Token): Expression {
+        return RpgParserFacade().createParser(
+            BOMInputStream(
+                ("".padStart(8) + token.text.substring(
+                    1,
+                    token.text.lastIndex
+                )).byteInputStream(Charsets.UTF_8)
+            ), LinkedList(), longLines = true
```
ftomassetti:
```suggestion
            ), errors = mutableListOf(), longLines = true
```

This could be slightly more obvious

ftomassetti:
Regarding your question: square brackets are added during parsing, in order to reuse the basic parser for expressions.

You can see how we do that by looking at RpgParserFacade.preprocess

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/MuteExecutionTest.kt
f-lombardo: They were the same test we have in InterpreterTest class

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
```diff
@@ -245,6 +249,17 @@ class InternalInterpreter(val systemInterface: SystemInterface) {
                 is MuteTimeoutAnnotation -> {
                     // Skip
                 }
+                is MuteFailAnnotation -> {
+                    val message = interpretConcrete(it.message)
+                    log(MuteAnnotationExecutionLogEntry(this.interpretationContext.currentProgramName, it, message))
```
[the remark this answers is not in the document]
f-lombardo: I will add a bit more specific log, but I thing we don't need a too detailed entry

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/mute.kt
f-lombardo: You are right, but adding a nullability would add a mess of null checking syntax in the code.
So I created an abstract class

docs/mute.md
f-lombardo: OK

docs/mute.md
f-lombardo: OK

docs/mute.md
f-lombardo: Good point: I modified the docs

rpgJavaInterpreter-core/src/test/resources/mute/SIMPLE_MUTE_FAIL_EVALUATED_MESSAGE.rpgle
f-lombardo: OK

rpgJavaInterpreter-core/src/test/resources/mute/SIMPLE_MUTE_FAIL_STATIC_MESSAGE.rpgle
```diff
@@ -0,0 +1,7 @@
+     DA                S              8  0 INZ(5)
+     DB                S              8  0 INZ(8)
+     D RESULT          S              8  0 INZ(0)
+      *
+    MU* FAIL['This code should not be executed']
+     C                   EVAL      RESULT = A + B
+     C                   SETON                                        LR
```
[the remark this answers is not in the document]
f-lombardo: OK

PR #14, opened 2019-10-25
Data package merge 1

ftomassetti:
Second delivery of data package.

Now, we still need to refine the contents of this package, however we wanted to have this merged quickly because we perform significant refactorings and we want:

1) The rest of the team to start benefit from these
2) Avoid conflicts on both sides

We plan to do a follow up to address several styling issues and we have already a list of those points: from comments to remove, to blank lines to add, to minor refactorings.

We will do those fixes as follow ups.

We are also aware that PRs should be way smaller. This was a particular case as we are changing how some core structures work. However we should strive to deliver much smaller changes in the future.


ftomassetti:
Thank you @benetti-smeup !

PR #17, opened 2019-10-28
Added timeout annotation to performance tests

benetti-smeup:
Added TIMEOUT annotation to performance tests. Renamed MUTE10_04_A in MUTE10_04A (max source name length 10 char). Removed MUTE11_02 from performance tests.


rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/MUTEExamplesTest.kt
```diff
@@ -20,13 +20,8 @@ class MUTEExamplesTest {
     }
 
     @Test @Category(PerformanceTest::class)
-    fun executeMUTE10_04_A_perf_strings() {
-        assertEquals(listOf("001.1_d01.1_A01.1_c01.1_B01.1_b01.1_C01.1_901.1_101."), outputOf("performance/MUTE10_04_A"))
-    }
-
-    @Test @Category(PerformanceTest::class)
```
ftomassetti: Why this test has been removed?

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/MUTEExamplesTest.kt
```diff
@@ -20,13 +20,8 @@ class MUTEExamplesTest {
...
-    fun executeMUTE11_02() {
```
ftomassetti: Why this test has been removed?

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/MUTEExamplesTest.kt
benetti-smeup: This test has been renamed, not removed.
MUTE10_04_A is not a valid source name (longer than 10 char) so I renamed it in MUTE10_04A.

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/MUTEExamplesTest.kt
benetti-smeup: MUTE11_02 is not a performance test. Is a program to test integration in the Smeup Gateway.

f-lombardo: Marco, that's true, but maybe we could keep it as a performance test too, don't you think?

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/MUTEExamplesTest.kt
f-lombardo: In this case the modification is so small that I thing it's worth it, anyway I think that, in the interpreter world, we could accept program names with longer names, especially if they are test programs

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/MUTEExamplesTest.kt
benetti-smeup: But it's a program that was created to run within the gateway as a folder listener. I don't think it can be useful as a performance test. It can't run correctly.

f-lombardo: OK, good point

ftomassetti:
Given all tests pass now I can approve it

PR #18, opened 2019-10-28
implemented SUB opcode

benetti-smeup:
Implemented SUB opcode and related MUTE13_06.

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
@@ -382,6 +382,22 @@ class InternalInterpreter(val systemInterface: SystemInterface) {
                     }
                     assign(statement.target, value.negate())
                 }
+                is SubStmt -> {
+                    val minuend = if (statement.factor1 == null) {
+                        interpret(statement.target)
+                    } else {
+                        interpret(statement.factor1)
+                    }
+                    require(minuend is NumberValue) {
+                        "$minuend should be a number"
+                    }
+                    val subtrahend = interpret(statement.factor2)
+                    require(subtrahend is NumberValue) {
+                        "$subtrahend should be a number"
+                    }
+                    val newValue:DecimalValue = DecimalValue(minuend.asDecimal().value.subtract(subtrahend.asDecimal().value))
+                    assign(statement.target, newValue)
+                }
                 is TimeStmt -> {
                     when (statement.value) {
                         is DataRefExpr -> {
@@ -1280,6 +1296,7 @@ class InternalInterpreter(val systemInterface: SystemInterface) {
         return dataDefinition.type.blank()
     }
 }
+
 
 private fun AbstractDataDefinition.canBeAssigned(value: Value): Boolean {
     return type.canBeAssigned(value)
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
@@ -24,6 +24,7 @@ abstract class Value {
 
 interface NumberValue {
     fun negate(): Value
+
 }
 
 // TODO Should we change value to a val in order tho share instances?
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
@@ -262,6 +262,21 @@ data class ZSubStmt(
         return emptyList()
     }
 }
+data class SubStmt(
+    val factor1: Expression?,
+    val target: AssignableExpression,
+    @Derived val dataDefinition: InStatementDataDefinition? = null,
+    val factor2: Expression,
+    override val position: Position? = null
+) :
+    Statement(position), StatementThatCanDefineData {
+    override fun dataDefinition(): List<InStatementDataDefinition> {
+        if (dataDefinition != null) {
+            return listOf(dataDefinition)
+        }
+        return emptyList()
+    }
+}
 
 data class TimeStmt(
     val value: Expression,
--- a/mute_tests/MUTE13_06.rpgle
+++ b/mute_tests/MUTE13_06.rpgle
@@ -1,0 +1,81 @@
+   COP* *NOUI
+     V*=====================================================================
+     V* MODIFICHE Ril.  T Au Descrizione
+     V* gg/mm/aa  nn.mm i xx Breve descrizione
+     V*=====================================================================
+     V* 28/10/19  001221  BMA   Creazione
+     V*=====================================================================
+     V* OBIETTIVO
+     V*  Programma finalizzato ai test sulla SUB
+     V*
+     V*=====================================================================
+     D*----------------------------------------------------------------
+     D AAA005          S              5
+     D AAA010          S             10
+     D DSP             S             50
+      *
+     D  N1             S              9  0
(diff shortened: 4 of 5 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
```diff
@@ -382,6 +382,22 @@ class InternalInterpreter(val systemInterface: SystemInterface) {
                     }
                     assign(statement.target, value.negate())
                 }
+                is SubStmt -> {
+                    val minuend = if (statement.factor1 == null) {
```
ftomassetti: Factor1, factor2, and target are not very clear names.
In my opinion we should determine the minuend and subtrahend when performing the translation from the parse tree to the ast.
At this point we should simply access them from the AST node, not performing this logic

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
```diff
@@ -382,6 +382,22 @@ class InternalInterpreter(val systemInterface: SystemInterface) {
...
+                        interpret(statement.factor1)
+                    }
+                    require(minuend is NumberValue) {
+                        "$minuend should be a number"
+                    }
+                    val subtrahend = interpret(statement.factor2)
+                    require(subtrahend is NumberValue) {
+                        "$subtrahend should be a number"
+                    }
+                    val newValue:DecimalValue = DecimalValue(minuend.asDecimal().value.subtract(subtrahend.asDecimal().value))
```
ftomassetti: it is not necessary to specify the type of newValue, as it can be inferred.

Regarding the value, I am not sure we always want to get a DecimalValue. In most languages when subtracting two integers I obtain an integer. I obtain a decimal value when at least one of the values is a decimal.
This is important because if the target is an integer trying to assign a DecimalValue should trigger an error (unless coerce fix that for us, but we need to be sure this is the behavior we want)

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -24,6 +24,7 @@ abstract class Value {
 
 interface NumberValue {
     fun negate(): Value
+
```
ftomassetti: I think we should remove this extra blank line

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
```diff
@@ -1281,6 +1297,7 @@ class InternalInterpreter(val systemInterface: SystemInterface) {
     }
 }
 
+
```
ftomassetti: I think we should remove this extra blank line

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -262,6 +262,21 @@ data class ZSubStmt(
         return emptyList()
     }
 }
+data class SubStmt(
```
ftomassetti: I think we should add a new blank line before SubStmt

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -262,6 +262,21 @@ data class ZSubStmt(
...
+    val factor1: Expression?,
```
ftomassetti: I would use more expressive names then factor1 and factor2, like subtrahend and minuend

ftomassetti: Also "left" would be a better name, in my opinion

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
benetti-smeup: but minuend can be factor 1 if  factor 1 is not blank or result if factor 1 is blank.
it's possible to perform this logic translating from the parse tree to the ast?

ftomassetti: yes, it is possible. There we can write all logic we desire using Kotlin.
The idea is to "normalize" the AST, making our work as easy as possible in the interpreter.

The parse tree (i.e., the context classes) is complicated because it derives directly from what we parse. The AST can be simpler and we can perform the simplification during the translation from the parse tree to the AST.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
benetti-smeup: How can i do that? the Value class has not a subtract method. I had to cast minuend. Is there another way?

ftomassetti: Using the `is` operator. Look at what we do for the PlusExpr, for example:

```
            is PlusExpr -> {
                val left = interpret(expression.left)
                val right = interpret(expression.right)
                when {
                    left is StringValue && right is StringValue -> {
                        val s = left.valueWithoutPadding + right.valueWithoutPadding
                        StringValue(s)
                    }
                    left is IntValue && right is IntValue -> IntValue(left.value + right.value)
                    else -> throw UnsupportedOperationException("I do not know how to sum $left and $right at ${expression.position}")
                }
            }
```

Also, in this way you can check we are subtracting things with the proper type. We do not want to allow subtracting two strings, for example.

ftomassetti: Also, if you want you could add this logic in a subtract method for Value

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
benetti-smeup: Done

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
benetti-smeup: done

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
benetti-smeup: done

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
benetti-smeup: done

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
benetti-smeup: implemented with when

benetti-smeup:
Finished, i think.

benetti-smeup:
this code is included in statementADD1 branch

PR #31, opened 2019-11-06
Mute timeout fix

f-lombardo:
About the problem of estimating interpreter performances


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/mute/muterunner.kt
```diff
@@ -120,9 +120,12 @@ fun executeWithMutes(
             sorted.forEach { (line, annotation) ->
                 if (verbose || !annotation.succeeded()) {
                     println("Mute annotation at line $line ${annotation.resultAsString()} - ${annotation.headerDescription()} - ${file.linkTo(line)}".color(annotation.succeeded()))
-                    if (annotation is MuteComparisonAnnotationExecuted && !annotation.succeeded()) {
-                        println("   Value 1: ${annotation.value1Expression.render()} -> ${annotation.value1Result}")
-                        println("   Value 2: ${annotation.value2Expression.render()} -> ${annotation.value2Result}")
+                    if (!annotation.succeeded()) {
```
ftomassetti: I would consider adding a property named failed in annotation. I think it would make this slightly more readable

f-lombardo: I agree

PR #32, opened 2019-11-06
Fl03

f-lombardo:
PR about smoothing the usage of the interpreter by external users


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
```diff
@@ -1389,17 +1392,15 @@ class InternalInterpreter(val systemInterface: SystemInterface) : InterpreterCor
                     return StringValue(sourceString.replaceRange(0..replStringLength - 1, replString))
                 }
                 // case of %REPLACE(stringToReplaceWith:stringSource:startIndex)
-                if (expression.start != null && expression.length == null) {
+                if (expression.length == null) {
```
ftomassetti: Isn't a risk start being null?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/utils/misc.kt
```diff
@@ -70,6 +70,16 @@ fun String?.asDouble(): Double {
         return 0.0
     }
 }
+fun String?.asBigDecimal(): BigDecimal? {
```
ftomassetti: I would put a blank line before this one.
Also, why supporting `String?` instead of `String` if anyway we want to return null when the string is null?
One could support just string and then use it as `myStringValuePossiblyNull?.asBigDecimal()`. This would result in a bit less code in this function

f-lombardo: Oh, yes, good idea!

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
f-lombardo: The previous lines are
`               if (expression.start == null) {
                    return StringValue(sourceString.replaceRange(0..replStringLength - 1, replString))
                }`
So the compiler suggests to remove the null check

PR #50, opened 2019-12-03
Added forAll and used in slow tests

### LINKED ISSUE #46: Tag slow tests as @PerformanceTest and add some property tests instead
Some unit tests, such as `RpgDataEncoderTest`, are slow. Mark them as `@PerformanceTest`and and create some equivalent tests that use a "probabilistic" approach, i.e. tryng some random examples. It could be useful to execute them with some library for property testing like [kwik](https://github.com/jcornaz/kwik)

f-lombardo:
## Tag slow tests as @PerformanceTest and add some property tests instead
Closes #46

### DIFF
--- a/rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/overlay/RpgDataEncoderTest.kt
+++ b/rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/overlay/RpgDataEncoderTest.kt
@@ -1,7 +1,11 @@
 package com.smeup.rpgparser.overlay
 
+import com.smeup.rpgparser.PerformanceTest
 import com.smeup.rpgparser.interpreter.*
+import com.smeup.rpgparser.test.forAll
+import com.smeup.rpgparser.test.integerGenerator
 import org.junit.Test
+import org.junit.experimental.categories.Category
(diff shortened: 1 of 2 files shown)


f-lombardo:
private val encodeDecodePackedLambda: Int.() -> Boolean 

It's a lambda with receiver https://medium.com/tompee/idiomatic-kotlin-lambdas-with-receiver-and-dsl-3cd3348e1235

Here we declare a val that is a function which gets an integer parameter and gives back a Boolean. Since it's 
`Int.() -> Boolean`, i.e. the `Int` parameter is outside (),  it's seen as `this` inside the lambda (it's what is called the receiver).`this`can be omitted, so `toBigDecimal`can be called directly

f-lombardo:
> I don't understand the purpose of this implementation.
> Can you explain it, please?

The idea is this: there are some tests that are really slow, because they run for a great amount of numbers (-999999..999999). Let's mark these tests as `performancetest `, so that they run in hour in-house jenkins. In order to keep a safe guard, though, let's run these test for a smal subsample of numbers in that interval, say 100 random numbers. This is the purpose of the `forAll` pseudo-statement that I "borrowed" :-) from other libraries, for example https://github.com/jcornaz/kwik

PR #80, opened 2020-01-09
Added test on array of unequal size

benetti-smeup:
The test is ignored for an error parsing an array's element as result of movel


f-lombardo:
OK, thanks

PR #129, opened 2020-03-17
Fixes, partial implementation of lookup, refactoring of internal_interpreter

f-lombardo:
Fixes, partial implementation of lookup, refactoring of internal_interpreter


f-lombardo:
Rì

> Why rpgJavaInterpreter-core/src/test/resources/LOOKUP_OP2.rpgle and rpgJavaInterpreter-core/src/test/resources/LOOKUP_OP3.rpgle have 2 statements with SETON LR?

Oh, thank you: fixed!

PR #152, opened 2020-04-28
Improving EDITC

f-lombardo:
- Improving design using enum class for decedit
- Minor performance optimzation create less DecimalFormatSymbols and reducing useless instructions when logging is off


f-lombardo:
> 
> 
> I think that localization proprieties such as Cp037 (code page for EN_US) or Locale.ITALY should be inferred from a external configuration and centralized and not hardcoded in the sources.

Indeed they are: we are just optimizing the editc for those use case, that are the most used ones

PR #166, opened 2020-05-07
Symboltable optimization

lanarimarco:
## Description
A gentle but  no trivial solution to optimize SymbolTable.
Introduced key consepts:
- FeaturesFactory
- MainExecutionContext

### FeaturesFactory
It lets to try new features in jariko via property file o system property.
SymbolTable now is an interface (`ISymbolTable`).
SymbolTable implementation is a feature provided by `IFeaturesFactory `(also FeaturesFactory is an interface), so, via configuration you can enable standard SymbolTable or experimental SymbolTable, by choose of the appropriate FeaturesFactory implementation.

### MainExecutionContext
It lets to propagate attributes (`mapOf<String, Any>`) in any class, instance or function, the scope is current thread (main current execution program).
This one in the future, could be interesting  when we should deal with the activation group and  more in general to solve some stuff about memory scope handling, local, global etc...


rpgJavaInterpreter-core/build.gradle
```diff
@@ -248,4 +252,44 @@ for further informations."""
     }
 }
 
-check.dependsOn runMutes
+task profileMainClass (type: JavaExec) {
```
f-lombardo, line 255: I think there is a small mistake here: the check.dependsOn runMutes should not be deleted

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/IFeaturesFactory.kt
```diff
@@ -0,0 +1,41 @@
...
+}
+
+object FeaturesFactory {
+
+    private val factory: IFeaturesFactory by lazy {
+        val property = System.getProperty("featuresFactory", "")
+        val featuresFactoryId = if (property == "") "default" else {
+            property
+        }
+        val featuresFactoryImpl = if (featuresFactoryId.indexOf(".") != -1) {
```
f-lombardo, line 21: Maybe 
`featuresFactoryId.contains(".")`

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/ISymbolTable.kt
```diff
@@ -0,0 +1,45 @@
+package com.smeup.rpgparser.interpreter
+
+interface ISymbolTable {
```
f-lombardo, line 3: I think that adding an I in front of an interface couples too much to a particular design choice. What if we want to use an abstract class later? Obviously this is just a matter of taste, so every choice is a good one :-)

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/SymbolTable.kt
```diff
@@ -51,27 +51,14 @@ class SymbolTable {
...
             values.keys.firstOrNull { it.name.equals(dataName, ignoreCase = true) }
 
-    operator fun set(data: AbstractDataDefinition, value: Value) {
+    override operator fun set(data: AbstractDataDefinition, value: Value): Value? {
         require(!(data !in this && data.name in this)) {
             "This data definition would conflict with an existing data definition with the same name. This data definition: $data. Existing data definition: ${this[data.name]}"
         }
         require(data.type.canBeAssigned(value))
-        values[data] = value.forType(data.type)
+        return values.put(data, value.forType(data.type))
```
f-lombardo, line 62: `values[data] =` is a bit more kotlinish

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/program.kt
```diff
@@ -16,6 +16,9 @@ interface Program {
 }
 
 class RpgProgram(val cu: CompilationUnit, dbInterface: DBInterface, val name: String = "<UNNAMED RPG PROGRAM>") : Program {
+
+    private var interpreter: InternalInterpreter? = null
```
f-lombardo, line 20: Maybe a val with by lazy initialization would be better, avoiding !! calls

rpgJavaInterpreter-core/build.gradle
lanarimarco: You are right.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/IFeaturesFactory.kt
lanarimarco: Ok

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/ISymbolTable.kt
lanarimarco: We could talk about it in the next future, but I agree that my choice could be not the best choice :-)

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/SymbolTable.kt
lanarimarco: I need to return the old value but "values[data]" is not expression.
Have you some suggestion?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/program.kt
lanarimarco: InternalInterpreter requires a non optional SystemInterface instance.
I thought to proceed as follows:

```
class RpgProgram(val cu: CompilationUnit, dbInterface: DBInterface, val name: String = "<UNNAMED RPG PROGRAM>") : Program {

    private var systemInterface: SystemInterface? = null

    private val interpreter: InternalInterpreter by lazy {
        InternalInterpreter(this.systemInterface!!)
    }
```

and to change execute implementation like this:
```
    override fun execute(systemInterface: SystemInterface, params: LinkedHashMap<String, Value>): List<Value> {
        require(params.keys.toSet() == params().asSequence().map { it.name }.toSet()) {
            "Expected params: ${params().asSequence().map { it.name }.joinToString(", ")}"
        }
        this.systemInterface = systemInterface
```

This approach works, but I think it may be interesting for the future, to expose SystemInterface through MainExecutionContext.
In this way we are sure that SystemInterface will be immutable within its execution context

What do you think?

f-lombardo: I have to consider this. At the moment I think your solution is pretty good.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/SymbolTable.kt
f-lombardo: You're right

PR #175, opened 2020-09-03
Parm fix

f-lombardo:
Variables defined inside C specs (for example PARM specs) where not added to the list of valid variables if their definition was nested in block of statements related to IF/FRO/DO/SELECT


lanarimarco (approved):
OK

PR #177, opened 2020-09-09
StringValue is assignable to CharacterType

f-lombardo:
Added an RPG example of data conversion between strings and DS subfield involving StringValue and CharacterType


lanarimarco (approved):
OK

PR #178, opened 2020-09-09
DS tests and solutions

f-lombardo:
## DS related issues

- A longer DS is assignable to a shorter one
- Added failing test on DS field offset calculation (TO SOLVE!!!)
- Clear for DS fields


lanarimarco (approved):
OK

PR #180, opened 2020-09-10
Fix on DS fields defined with start/end position

f-lombardo:
The end position is no more interpretated ad field precision if there exists a start position definition like in

`     D DSTEMP          DS            64
     D VAIDOJ                  1     10A
     D VAATV0                 11     11A`

PR #181, opened 2020-09-15
Cumulative PR - "Fix on DS fields defined with start/end position" - "Fix on data definition with C specs"

f-lombardo:
#180 Fix on DS fields defined with start/end position 
The end position is no more interpretated ad field precision if there exists a start position definition like in
D DSTEMP DS 64 
D VAIDOJ 1 10A 
D VAATV0 11 11A

#179 Fix on data definition with C specs
Statements that define data can now be positioned after their first usage


ftomassetti (approved):
Mi sembra in linea con le altre PR e la discussione avuta su quelle. Mi sembra quindi una buona soluzione

PR #193, opened 2020-10-08
Fixed a potential issue that can occur if the same program is called twice and others little interesting things

lanarimarco:
## Description
 * Fixed a potential issue that can occur if the same program is called three or more times and the first time comes out in RT while the others comes out in LR. On the third call, the symbol table is not initialized correctly.

 * Added JarikoCallback.onEnterPgm and JarikoCallback.onExitPgm in order to inspect  memory, before and after program execution.

 * Improved and added some unit test.

## Checklist:


f-lombardo (approved):
OK. Just as a side-note, on AS400 Activation Groups are case-insensitive

PR #194, opened 2020-10-09
Time duration

f-lombardo:
## Implementing SUBDUR and %DIFF for *DAYS duration

First draft of a possible implementation of operations that give the difference in days between two dates


lanarimarco (approved):
Ok

PR #196, opened 2020-10-16
Draft version of how to pass a DS to Jariko.

lanarimarco:
## Description

Draft version of how to pass a DS to Jariko.

## Checklist:
- [] There are tests regarding this feature
- [] There is a specific documentation in the `docs` directory

### DIFF
--- a/examples/src/main/kotlin/com/samples/jariko/RpgSamples.kt
+++ b/examples/src/main/kotlin/com/samples/jariko/RpgSamples.kt
@@ -4,10 +4,12 @@ import com.smeup.rpgparser.execution.Configuration
 import com.smeup.rpgparser.execution.JarikoCallback
 import com.smeup.rpgparser.execution.executePgmWithStringArgs
 import com.smeup.rpgparser.execution.getProgram
-import com.smeup.rpgparser.interpreter.ISymbolTable
+import com.smeup.rpgparser.interpreter.*
 import com.smeup.rpgparser.jvminterop.JavaSystemInterface
+import com.smeup.rpgparser.parsing.parsetreetoast.RpgType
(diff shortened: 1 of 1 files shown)


examples/src/main/kotlin/com/samples/jariko/RpgSamples.kt
```diff
@@ -68,4 +70,58 @@ fun execWithCallBack(programSource: String, programArgs: List<String>, jarikoCal
...
+    dataStructValue.set(
+        field = FieldDefinition(
+            name = "Surname",
+            type = StringType(5, false),
+            explicitStartOffset = 5,
+            explicitEndOffset = 0
+        ),
+        value = StringValue("Smith")
+    )
+    // note rpgType = RpgType.ZONED is always required regardless of number type definition in rpg program
```
benetti-smeup, line 117: Why should I use RpgType.ZONED for all kind of numeric type?

lanarimarco: This is a very long story. I will try to explain to you.
Jariko only accepts a list of strings as input parameters.
First of all I had to find a way to represent a DS as a string.
I found that DataStructValue (the DataStruct object model in Jariko) is able to provide a good string representation of itself, via the DataStructValue.value function, only for numeric fields defined as RpgType.ZONED.
However we are only talking about a draft version, and these problems will be addressed in a deliverable, in fact this is not the solution but a possible solution.

benetti-smeup: ok, thanks

benetti-smeup (approved):
ok

PR #204, opened 2020-11-05
Add multithreading test execution

parodi-smeup:
## Description

Include a summary of the change.

Related to # (issue)

## Checklist:


rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/interpreter/SymbolTableStoragingTest.kt
```diff
@@ -347,4 +348,50 @@ class SymbolTableStoragingTest {
...
+            for (testName in testNames) {
+                var simbolTableStoragingTest = SymbolTableStoragingTest()
+                val workerThread: Runnable = WorkerThread(simbolTableStoragingTest, testName)
+                executor.execute(workerThread)
+            }
+        }
+        executor.shutdown()
+
+        println("Waiting for all thread to finish...")
+        while (!executor.isTerminated) {
```
lanarimarco, line 373: A very little change request. It's better to use: Executor.awaitTermination with a reasonable timeout (60 secs) in order to avoid potentially infinite loop in case that executor.isTerminated never return true

lanarimarco (changes requested):
Little change...

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/interpreter/SymbolTableStoragingTest.kt
parodi-smeup: It sounds good

lanarimarco (approved):
Good!!!

PR #205, opened 2020-11-06
New programs for testing parsing performance

benetti-smeup:
New programs with different samples of entry


parodi-smeup:
Ok

parodi-smeup (approved):
Too cool to be true

PR #207, opened 2020-11-12
Improvement of unit tests on parser performance

lanarimarco:
Added a new "parsing" channel with the purpose to log only AST creation performance.
Now we can pass --lc parameter also to standalone mute runner in order to log every mutes via cli.
Released MutePerformanceAstTest in order to verify inside jenkins the performance evolution trend only for parsing phase
Added a new set of mutes intended to test parsing performance

## Checklist:


benetti-smeup (approved):
ok

PR #214, opened 2020-12-02
Mutes serialized ast

lanarimarco:
## Description

Now, serialization coverage is for all test cases.

All tests for wich the rpg sources are passed to jariko through theirs name have been duplicated in order to execute the same tests using either rpg compiled than rpg interpreted.

Introduced a dependency from all rpg files compilation in "gradle check".
If compilation fails then "gradle check" fails, this is warranty that the tests case using compiled rpgs, are using really the compiled version.

## Checklist:
- [] There is a specific documentation in the `docs` directory

### DIFF
--- a/rpgJavaInterpreter-core/build.gradle
+++ b/rpgJavaInterpreter-core/build.gradle
@@ -297,4 +297,9 @@ for further informations."""
     }
 }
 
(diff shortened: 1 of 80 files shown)


benetti-smeup (approved):
ok

PR #201, opened 2020-11-03
Fix defaultactgrp

lanarimarco:
## Description

The default activation name is passed through a non-mandatory property: Configuration.defaultActivationGroupName.

## Checklist:
- [X] There are tests regarding this feature
- [X] The code follows the Kotlin conventions (run `./gradlew ktlintCheck`)
- [X] The code passes all tests (run `./gradlew check`)
- [] There is a specific documentation in the `docs` directory

PR #217, opened 2020-12-03
Fixed symboltable serialization error

lanarimarco:
Fixed symbol table serialization issue that occurred if caller and called programs was running in activation groups
Regression introduced with the pull request #201.
Added unit test X1X2110 used to reproduce error solved by this commit.
Added in configuration the callback function that allows to obtain the AST (instance of CompilationUnit) before the interpretation, this function could be used to make more friendly the DataStructValue creation by passing the DS name and fields name belonging to DS.
Added IMemorySliceStorage.createMemoryStorage useful to test symboltable serialization, without the needed to implement the interface.

## Checklist:


benetti-smeup (approved):
ok

PR #227, opened 2020-12-18
Reload integration

smeup:
## Description

Added some new tests and minor improvments on native DB and Reload integration.

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -609,11 +609,16 @@ object HiValValue : Value {
         if (other is HiValValue) 0 else 1
 
     override fun asString(): StringValue {
-        TODO("Not yet implemented")
+        return this.asInt().asString()
```
lanarimarco, line 612: We need talk about this thing.
Hi and low value are related to type, try to search "fun Type.hiValue()" in coercing.kt.
Maybe is necessary to schedule a meet also with rpg expert to investigate about these stuff.

lanarimarco (changes requested):
In general is good, but there is a matter not too clear for me about high and low value.

PR #237, opened 2021-02-03
New call handler example

lanarimarco:
## Description
Added a use case, not too far from the reality, of a Caller Program Handler implementation
## Checklist:
- [] There are tests regarding this feature


parodi-smeup (approved):
Don't need to have an rpgle program as interface to know "the contract", now this information can be obtained froma an URL.

PR #243, opened 2021-03-08
Fixed syntactic error in mute.

lanarimarco:
Now compilation dir is deleted before performances test, this way we are sure that in compilation dir we don't have compiled files that should not be there.

## Checklist:


benetti-smeup (approved):
ok

PR #246, opened 2021-03-12
/API directive implementation

lanarimarco:
## Description
Currently we can include /API in all programs but the API loading will always be Static, which it means that the API's CompilationUnit always will be included into program that declares the use of that API.

## Checklist:


parodi-smeup (approved):
Well done!

PR #247, opened 2021-03-16
Fixed issues about the missing of overflow check for sum, subtract an…

lanarimarco:
…d multiply operation, in case of int type variables.

## Checklist:


benetti-smeup (approved):
ok

PR #254, opened 2021-04-07
Fixed reload integration compilation issue.

lanarimarco:
Changed reload dependency from compile to api because compile dependency has been deprecated and if we want to expose reload as api to the projects dependant from jariko, api dependency is the right mode.

## Checklist:


benetti-smeup (approved):
ok

PR #258, opened 2021-05-05
Now executeMUTE15_10 is no longer ignored.

lanarimarco:
Improved mute failures logging when mute is executed through executePgm. If you pass `configurations.options.dumpSourceOnExecutionError property` set to `true`, the source program will be dumped in error stream in case of errors.

## Checklist:


parodi-smeup (approved):
:1st_place_medal:

PR #263, opened 2021-05-13
Fixed symboltable serialization error for programs using procedures

lanarimarco:
## Description
The problem occurs when Jariko is called by passing an IMemorySliceStorage instance via Configuration and then enabling serialization, and this is how jariko works in production environments.  
Error stacktrace is as follows:
```
java.lang.RuntimeException: persist property not set for these slices: 
[MemorySlice(memorySliceId=MemorySliceId(activationGroup=*DFTACTGRP, programName=<UNSPECIFIED>), symbolTable=com.smeup.rpgparser.experimental.SymbolTable@671083a7)] 
```

## Checklist:


parodi-smeup (approved):
Ok

PR #264, opened 2021-05-18
Bugfix/nullpointer on call parm variables

parodi-smeup:
## Description

This branch fix exception caused by a missing implementation of "PListParm" statement.
"PListParm" now extends "StatementThatCanDefineData", so it is able to declare/instantiate variables. 
The test 'InterpreterTest.executePGM_A()' is used to check the implementation works as expected.
A change in test 'RunnerTest.executeExampleWithCall()' was needed, related to assert. 

Related to # 'nullpointer exception' trying to get a variable declared by 'PARM' statement from 'SymbolTable'.

## Checklist:


lanarimarco (changes requested):
You should write just two words in pr comment and compiling properly the check list.

lanarimarco (approved):
Ok, the best (but is just a suggestion for next times) it could be to see a stack trace of the fixed bug.
Try to see [my pull](https://github.com/smeup/jariko/pull/263)

PR #269, opened 2021-07-26
Api's mute test

FiorenzaBusi:
Added 4 new mute and the relatives API.
Modified MuteExecutionTest because added 4 new test about mute execution.

## Description

Added mutes related the API directive

## Checklist:


parodi-smeup (changes requested):
Not a mandatory request change but only a best practice: MUTE18_04.rpgle should be use a smart way to get lower (or greater) value.
As showing in following snippet code, values are set in an array, then the array is ordered  by 'SORTA' operation code, so first array element is the lower one:
`d values          s              2s 0 dim(3)                       
 *                                                                 
c                   eval      values(1) = 25                       
c                   eval      values(2) = 18                       
c                   eval      values(3) = 43                          
c                   sorta     values                               
 *                                                                 
c                   seton                                        lr
`

parodi-smeup (approved):
Ok

PR #271, opened 2021-08-27
Fix lookup bif (when using start/end indexes parameters)

parodi-smeup:
## Description

Add properly handle of optional parameters in **%LOOKUP** Built In Function, used to limit search into array.

- the start parameter is used as start element index to search forward;
- the length parameter is used as "number of elements" upper limit to search;

This implementation is done to avoid lookup fails search if above parameters are specified.
Test **MUTE13_28.rpgle** check properly implementation

## Checklist:


lanarimarco (approved):
Ok

PR #274, opened 2021-09-02
Fix multiple array declaration

FiorenzaBusi:
Change the implementation of array's association:
- When the array's declaration is `**CTDATA`, array's association is by name.
- When the array's declation is without  `CTDATA`, array's association is by index.

## Description

Added 4 new mutes to test the multiple array declaration.
In the class `MuteExecutionTest.kt` added 4 new function to test  the new mutes.
In `misc.kt` changed the logic of substring and replace of the array's name, which it was at the end of program.
In fact, when the array contains `**CTDATA` , we leave the name of the array; otherwise we replace the content of the string with a blank content.
In the class `cu-components.kt`, added new function to return a compileTimeArray using index.
In the class `internal_interpreter.kt:` added new variable called index, that variable helps us to cont the number of time that we have an ArrayValue. This index is used when we have to obtain the element of the array (value), if it is an array declared with "**".
When the array is declared with `**CTDATA` the value (element of array)  is set using the name of array.

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
@@ -199,6 +199,7 @@ open class InternalInterpreter(
             dbFileMap.add(it)
         }
 
+        var index: Int = 0
         // Assigning initial values received from outside and consider INZ clauses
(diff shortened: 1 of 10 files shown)


lanarimarco (changes requested):
The pull request description it should show, first of all what it fixes and finally, the implementation details.

lanarimarco (approved):
Ok

PR #281, opened 2021-11-11
Fixed dependency from reload version.

lanarimarco:
## Description

We have removed a snapshot dependency from reload

## Checklist:


parodi-smeup (approved):
:1st_place_medal:

PR #284, opened 2021-12-03
Fixed shell issue with character £

lanarimarco:
## Description

On window platform, if you paste rpgle code containing char £ into the interactive shell, it always gets errors due to bad character encoding.


parodi-smeup (approved):
:1st_place_medal:

PR #285, opened 2021-12-14
Update log4j2 dep: CVE-2021-44228

lanarimarco:
## Description
Update log4j2 dep: CVE-2021-44228


parodi-smeup (approved):
:1st_place_medal:

PR #294, opened 2022-04-19
"Reload" library transitive dependency issue

lanarimarco:
## Description
The "reload" library dependency is not properly propagated to third party libraries that use Jariko

Related to # (issue)

## Checklist:


parodi-smeup (approved):
Too cool to be true

PR #297, opened 2022-05-12
Reduce errorlog verbosity

lanarimarco:
## Description
Now in case of errors, the dump of the rpg source content will be dumped only if `Options.dumpSourceOnExecutionError=true` 

## Checklist:


benetti-smeup (approved):
ok

PR #306, opened 2022-07-27
Fix error when the factor 2 in opcode DO is not specified.

lanarimarco:
## Description
Jariko threw this error: factor2.content must not be null

## Checklist:


mattiabonardi (approved):
Ok !

PR #315, opened 2022-11-07
Feature/add before interpreation callback

lanarimarco:
## Description
Added onInterpreterCreation callback in order to make InterpreterCore available, this functionality is meant for the future because through InterpreterCore we can get low level access to Jariko and thus simplify the implementation of new features, for example, when Jariko is seen as debuggee.
Fix an assignment issue where was allowed to assign to a variable a number with precision greater than the variable itself.

## Checklist:


mattiabonardi (approved):
Ok!

PR #325, opened 2023-03-24
Feature/nw23000217 parms status support

lanarimarco:
### Main changes

Added support of `*PARMS` and `*STATUS` keywords when used in DS field declaration, `*STATUS` is always set to 0.
These two new keywords are modeled in `DSFieldInitKeywordType`, therefore, if we want to add new keywords at DS field level, we should add a new enum and implementing the right behavior based on what we have to do.

### Minor changes

- Fixed problem where `DIM(argument)` if specified into a DS, the `argument` was evaluated only if constant
- Added` jariko.compilePerformanceMutes` system property (default `true`). You can set this property to `false` in order to ovoid the compilation before the performance mutes test execution.
- Added `ParsingProgram.attributes` as map at scope `ParsingProgram`, at parsing time you can access to this property via `MainExecutionContext.getParsingProgramStack().peek().attributes`

## Checklist:


mattiabonardi (approved):
Ok!

PR #338, opened 2023-07-19
Bugfix/nw23000322 fix like def in parm

mossini-smeup:
## Description

Fixed the usage of `LIKE DEFINE` whose factor 1 is not a D specification but an inline variable defined in `PARM` inside a subroutine.

## Checklist:


lanarimarco (approved):
There's some duplicate code but we're talking about three or four lines of code, so OK.

PR #358, opened 2023-11-22
Bugfix/nw23001440/fix move and movel

mattiabonardi:
## Description

Refactor `MOVE` and `MOVEL` statement to easy support all data types. Move the `move` function from `move_a.kt` to `movel.kt` file because `move` and `movel` functions have the same logic. Support `DecimalValue` and `ArrayValue` for each functions. Restore the `coerce` function and removed it from `move` and `movel` because the Ibmi RPGLE data type to String generate different format. `MOVE` and `MOVEL` using their own string format. 

Related to # (issue)

## Checklist:


lanarimarco (changes requested):
## Refacor

An heavy refactor like this is accepted just when the purposes are clear.

You: Refactor `MOVE` and `MOVEL` statement to easy support all data types.
Me: Why?

You: Move the move function from `move_a.kt` to `movel.kt`
Me: Why? Have you tried to understand the meaning of these file? `move_a.kt` contains the logic used by the `MoveStmt` and MoveAStmt, the `movel.kt` of the `MoveLStmt.`
This has been a specific design choice, and any change must be justified.

## What are you fixing
Can you give me some reference of the sections and steps in MULANGT90?

mattiabonardi:
You: Refactor MOVE and MOVEL statement to easy support all data types.
Me: Why?
The previous code didn't perform a difference between character (MOVE: from right to left, MOVEL: from left to right) but an assignment is made.

```rpgle
 * previous
C                 MOVE          STR2         STR4           ->  MOVE ('AA', 'XXXX' -> 'AA') WRONG
C                 MOVEL         STR2         STR4           ->  MOVEL ('AA', 'XXXX' -> 'AA') WRONG
 * actual
C                 MOVE          STR4         STR4           -> MOVE ('AA', 'XXXX' -> 'XXAA') CORRECT
C                 MOVEL         STR4         STR4           -> MOVE ('AA', 'XXXX' -> 'AAXX') CORRECT
```

With refactoring each data type is treated the same way as if it were a string. Previously erroneous moves are performed for each given type.

You: Move the move function from `move_a.kt` to `movel.kt`
Me: Why? Have you tried to understand the meaning of these file? `move_a.kt` contains the logic used by the MoveStmt and MoveAStmt, the movel.kt of the MoveLStmt.

In RPGLE language `MOVE` and `MOVEL` are the same operation code but work with different alignment:
- `MOVE`: align variable to right and overwrite characters from `Factor 2` to `Result`
- `MOVEL`: align variable to left and overwrite characters from `Factor 2` to `Result`

`MOVEA` is similar but It is used only for Arrays. They could be all three together or all three separate.

You: Can you give me some reference of the sections and steps in MULANGT90?
I'm fixing delta errors of MOVEL but I realized that the `coerce()` method was not correct for what I was doing.

lanarimarco:
> You: Can you give me some reference of the sections and steps in MULANGT90? I'm fixing delta errors of MOVEL but I realized that the `coerce()` method was not correct for what I was doing.

Why are you not adding in jariko test cases, the steps extracted from MULANGT90 as I had suggested?
If you had followed my advise, we will have fixed the issue earlier?

mattiabonardi:
The errors are not from `MULANGT90` they are delta errors from `MULANGT10`

lanarimarco (approved):
Ok, next time remember, as much as possible, to refactor the code in a single commit.

Good work!

PR #359, opened 2023-11-23
fix: handle MOVEL(P) with varying variables

mattiabonardi:
## Description

Handle multiple `MOVEL(P)` with varying variables. The test case is an extract of `MULANGT10.rpgle` file.

Related to # (issue)

## Checklist:


lanarimarco (approved):
Ok

PR #365, opened 2023-11-28
feat: handle DESCEND keyword for Arrays

mattiabonardi:
## Description

Handle `DESCEND` keyword for Arrays. If Array is declared using `DESCEND` the `SORTA` statement sort array in descending mode.

Related to # (issue)
Fix error SORTA with DESCENT (NW23001440 - LS23095443)

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -1459,7 +1459,7 @@ data class ForStmt(
 @Serializable
 data class SortAStmt(val target: Expression, override val position: Position? = null) : Statement(position) {
     override fun execute(interpreter: InterpreterCore) {
-        sortA(interpreter.eval(target), interpreter.getLocalizationContext().charset)
+        sortA(interpreter.eval(target), target.type(), interpreter.getLocalizationContext().charset)
```
lanarimarco, line 1462: In kotlin fortunatly you can pass the parameter by using also their name, force himself to use this feature because is very helpful when, as you made in the sortA implementation, you add furthers parameters to the function signature

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -233,13 +233,17 @@ data class UnlimitedStringValue(var value: String) : AbstractStringValue {
  * Cp0280   EBCDIC ITALIAN
  * See: https://www.ibm.com/support/knowledgecenter/SSLTBW_2.1.0/com.ibm.zos.v2r1.idad400/ccsids.htm
  */
-fun sortA(value: Value, charset: Charset) {
+fun sortA(value: Value, type: Type, charset: Charset) {
```
lanarimarco, line 236: If you assume that the only type accepted is `ArrayType`, why have you defined the new parameter type as `Type`  and not `ArrayType?`.
If you allow to pass any instance of `Type`,  the first time that I use `sortA` and I pass for example `StringType,` we will have a `ClassCastException`

lanarimarco (changes requested):
Something to improve.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
mattiabonardi: Fixed

lanarimarco (approved):
Ok

PR #366, opened 2023-11-28
Bugfix/char bif trailng spaces

lanarimarco:
## Description

`%CHAR` fix
When the argument is a `StringValue`  instance, it trims anyway the result , also if `StringValue.varying` is `false`.

This change had an impact of all tests that makes comparisons between expected and actual value by using the `DSPLY` opcode.
In order to reduce as much as possible the impact of this change, I have changed the default behavior of some utility function.

## Checklist:


mattiabonardi (approved):
Ok!

lanarimarco:
Closed  because needs to rebase

PR #372, opened 2023-12-11
Bugfix/nw23001440/handle return stmt

mattiabonardi:
## Description

Set the RT indicator always on in `ReturnStmt` + test case.

If the program finish with `ReturnStmt` the `RT` indicator always set. If the program finish with `ReturnStmt` and before it the `LR` statement was set `LR` indicator wins (see kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt 943).

Related to # (issue)
MULANGT04 - SEZ_A20 - P02

## Checklist:


lanarimarco (approved):
Ok

PR #376, opened 2023-12-13
Fix/nw23001440/fix cat statement

mattiabonardi:
## Description

Fix `CAT` and `CAT(P)` statement. 

Errors:
- Content 2 of Factor 2 can be a Number or a Number Variable.
- If Factor 1 isn't defined, Factor 2 contains both string variable and number expression (string var`:`number expr) and result field has initial value, `CAT` statement concatenate factor 2, number of blanks to result field.

```rpgle
 * CAT(-,'XYZ':1) '99    ' -> '99 XYZ'
 C                   CAT       STR3:1        RES6
```
- If Factor 1 isn't defined, Factor 2 contains only the string variable and result field has initial value, `CAT` statement not concatenate factor 2 to result field.

```rpgle
 * CAT(-,'CDEF') 'AB    ' -> 'AB    '
C                   CAT       STR4          RES5
```

Related to # (issue)
MULANGT10 - SEZ_A70 - P02
MULANGT10 - SEZ_A70 - P03
MULANGT10 - SEZ_A70 - P04 (MULANGT90)
MULANGT10 - SEZ_A70 - P06
MULANGT10 - SEZ_A70 - P07
MULANGT10 - SEZ_A70 - P09
MULANGT10 - SEZ_A70 - P10

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
@@ -1515,9 +1515,9 @@ data class CatStmt(
         } else {
             result
         } else if (blanks != null) {
-            (factor1 + blanks + factor2)
+            (factor1.trim() + blanks + factor2)
         } else {
-            (factor1 + factor2)
+            (factor1.trim() + factor2)
         }
 
         interpreter.assign(target, StringValue(result))
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
@@ -1483,14 +1483,47 @@ data class CatStmt(
     val left: Expression?,
     val right: Expression,
     val target: AssignableExpression,
-    val blanksInBetween: Int,
+    val blanksInBetween: Expression?,
     @Derived val dataDefinition: InStatementDataDefinition? = null,
     override val position: Position? = null
 ) : Statement(position), StatementThatCanDefineData {
 
     override fun execute(interpreter: InterpreterCore) {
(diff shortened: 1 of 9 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -1483,49 +1483,43 @@ data class CatStmt(
     val left: Expression?,
     val right: Expression,
     val target: AssignableExpression,
-    val blanksInBetween: Int,
+    val blanksInBetween: Expression?,
```
lanarimarco, line 1486: It is not clear to me why the change of blanksInBetween from Int to expression, which however is nothing more that a computed number, has required so much changes in this function implementation.  

Have you try to replace the first statement from:
 `val blanksInBetween = blanksInBetween ` to
```
val blanksInBetween = blanksInBetween?.let { 
            interpreter.eval(it) 
        }?.asInt()?.value?.toInt() ?: 0 
```

lanarimarco (changes requested):
In my opinion too changes, maybe I am wrong, but you should  try to follow my suggestion.
Make me know.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
mattiabonardi: Because `blanksInBetween` is an Expression, not a value and in `parseTreeToAst` phase I don't know the value of the expression because the value may be change at runtime. In the previous version `CAT` statement don't accept a variable after `:` in factor 2.

lanarimarco (approved):
You should force yourself to use the scope functions and elvis operators, because they make your code less error prone.
Remember that in kotlin the variable type declaration is optional.
Let indellij helps you, for example by using of menu "Code -> Inspect code", in order to fix the warnings and to learn how to improve your code.

For the rest, great work for terrible opcode.

PR #381, opened 2023-12-21
fix: %REPLACE with varying

mattiabonardi:
## Description

Fix `%REPLACE` bif with varying strings.

**PS**: I used `+` operator for string concatenation because I think that in this case is more efficient than `StringBuilder`. There isn't iterations. If not, I will change it.

Related to # (issue)
MULANGT90 - T16_A20 - P02
MULANGT90 - T16_A20 - P05

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/ExpressionEvaluation.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/ExpressionEvaluation.kt
@@ -546,22 +546,36 @@ class ExpressionEvaluation(
     }
 
     override fun eval(expression: ReplaceExpr): Value {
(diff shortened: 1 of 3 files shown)


lanarimarco (approved):
I agree with what you say about StringBuilder.
Try this snippet and you will see that the first loop is faster than second one
```kotlin
fun main() {
    println(
        measureTimeMillis {
            for (i in 1..1000000) {
                val a = "aaaaaaaaaaaaa" + "bbbbbbbbbbbbb"
            }
        }
    )
    println(
        measureTimeMillis {
            for (i in 1..1000000) {
                val a = StringBuilder("aaaaaaaaaaaaa").append("bbbbbbbbbbbbb").toString()
            }
        }
    )
}
```

In my opinion StringBuilder is efficient when we need to concat in the same variable more values

PR #387, opened 2024-01-05
fix: %char format of the result of DivExpr

mattiabonardi:
## Description

Fix the format of the `%CHAR` expression when include a `DivExpr`.
After some testing I can confirm that the strange behavior described in the implementation is detected only in the case of the div expression.

Related to # (issue)
MULANGT04 - SEZ_A90 - P01 
MULANGT04 - SEZ_A90 - P03

## Checklist:


lanarimarco (approved):
Ok

PR #393, opened 2024-01-10
Bugfix/nw23001440/trim on char bif

cosentino-smeup:
## Description
In this branch we fixed the problem that occurred when we used %CHAR bif, with a variable containing spaces. The spaces were wrongly trimmed. There are few considerations to do:
- both string variables and string variables with VARYING parameter, will return spaces if the value of the variable contains them.
- VARYING string variables will be trimmed during concatenation operation if the value doesn't contain spaces.
- The "outputOf" method, mostly used in tests functions, allows you to trim spaces at the end of the results if you don't want to consider them during the test.

I removed the wrong "trim" invocations where they were not necessary: eval of CharExpr, and stringRepresentation.
The executeBIFCHAR test, shows you how to execute the test with either with or without "trimEnd" parameter.

Related to 
 - N/A

## Checklist:


lanarimarco (approved):
Ok

PR #394, opened 2024-01-10
Bugfix/nw23001440/count n not negative

foresti-smeup:
## Description

This bugfix resolve the "Count n not negative" error in execution of mute tests listed below. 

**After the correction, the reported error no longer occurs. However, the tests still fail because the implementation of the %CHAR function in Jariko behaves differently from that of the AS400 when the function receive an expression as input. Since %CHAR is supported only when receive a variable as input, the issue of delta in results has not been addressed. Nevertheless, it will be necessary to modify the MUTE tests on AS400 to remove the unsupported calls  of %CHAR function.**

Related to # (issue)

MULANGT90-T04_A70_P09 count n not negative
MULANGT90-T04_A70_P10 count n not negative

## Checklist:


lanarimarco (changes requested):
It is not clear to me what you have fixed.
The comment in rpgle program makes me confusion because the expected on AS400 is differrent by the actual (jariko).
Since that this delta error in T04_A70.rpgle do not help me to understand the problem, you should create a more understable test case too.

foresti-smeup:
Minimized the T04_10 test

lanarimarco (approved):
Ok

PR #395, opened 2024-01-12
Bugfix/nw23001440/char of decimal wrong format

cosentino-smeup:
## Description
In this branch has been fixed the Decimal format when printed with the %CHAR.

The following tests have been modified because of this fix:
- BIFCHAR
- T40_A10
- MUTE12_16: this test was actually wrong in the first place but the error didn't come up for some reason. I imported the source code on AS/400 and tested it to be sure that Jariko calculated the same result.

Related to: 
- T02_A50

## Checklist:


rpgJavaInterpreter-core/src/test/resources/data/ds/MUTE12_16.rpgle
```diff
@@ -86,7 +86,7 @@
     MU* VAL1(£JaxWT5) VAL2(12.345) COMP(EQ)
      C                   SETON                                        50
       *
-    MU* VAL1(£JaxWT6) VAL2(-0.12345678) COMP(EQ)
+    MU* VAL1(£JaxWT6) VAL2(-0.12345678) COMP(NE)
```
lanarimarco, line 89: This is not the solution, take in account that MU* stands for mute annotation, and it is exactly the value of that variable in as400

lanarimarco (changes requested):
Why MUTE12_16 change? I think it has nothing to do with the issue related to uninitialized number variables?

rpgJavaInterpreter-core/src/test/resources/data/ds/MUTE12_16.rpgle
cosentino-smeup: As I explained in the description, the test MUTE12_16 was wrong. The variable £JaxWT6 is equal to 0 and it's compared to the constant "-0.12345678". They can't be equal. Without this fix, the result was wrong. If you look at the Screenshot1 (branch develop), you will see that even if Jariko is comparing the different values (value1 = 0 and value2 = -0.12345678) returned true as result when was testing if the variables were Equal. The problem is that it was comparing an intValue with a DecimalValue. In the Screenshot2 (current branch) you can see that now there are 2 Decimal values and the test return true as result when it's testing if the variables are Different.

Screenshot1

Screenshot2

lanarimarco: Ok, but in this case a NE(argument) test is much less significant than an EQ(argument) test, because NE(argument)  means infinite values except the argument and then: Ok, this test is wrong, but make it better with EQ (like before) instead of NE.

cosentino-smeup: done

lanarimarco (approved):
Ok

PR #401, opened 2024-01-18
Bugfix/nw23001440/movea with in

cosentino-smeup:
## Description
In this branch has been fixed the implementation of *IN in MOVEA statements. It was already partially implemented but not working. A new word has been added to the lexer and parser: SPLAT_ALL_INDICATORS. The function eval for the GlobalIndicatorExpr, that before caused a runtime exception, has been properly implemented to return all indicators, from 1 to 99, with their current value.

Related to: N/A

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/ExpressionEvaluation.kt
```diff
@@ -676,8 +676,20 @@ class ExpressionEvaluation(
     override fun eval(expression: AssignmentExpr) =
         throw RuntimeException("AssignmentExpr should be handled by the interpreter: $expression")
 
-    override fun eval(expression: GlobalIndicatorExpr) =
-        throw RuntimeException("PredefinedGlobalIndicatorExpr should be handled by the interpreter: $expression")
+    override fun eval(expression: GlobalIndicatorExpr): Value {
```
lanarimarco, line 679: You are iterating 3 times in order to convert a map of 99 items. into a string of 0 and 1. 
I would have implemented this algorithm in this way:
```
    val value = StringBuilder()
        for (i in IndicatorType.Predefined.range) {
            value.append(if (interpreterStatus.indicators[i] == null) "0" else if (interpreterStatus.indicators[i]!!.value) "1" else "0")
        }
        return StringValue(value.toString())
```
I tried it and it is work but maybe I makes it too easy.
What do you think about it?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/indicators.kt
```diff
@@ -22,9 +22,7 @@ data class IndicatorExpr(val index: IndicatorKey, override val position: Positio
 @Serializable
 data class GlobalIndicatorExpr(override val position: Position? = null) :
         AssignableExpression(position) {
-    override fun size(): Int {
-        TODO("not implemented") // To change body of created functions use File | Settings | File Templates.
-    }
+    override fun size(): Int = 99
```
lanarimarco, line 25: Replace 99 with the right constant value which is: `IndicatorType.Predefined.range.last`

cosentino-smeup: Done

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/ExpressionEvaluation.kt
cosentino-smeup: I think you are right. It's an improvement of the performance and optimization of the code.

lanarimarco (approved):
Ok

lanarimarco:
Ok

PR #406, opened 2024-01-24
Bugfix/zero number literal initialization

lanarimarco:
## Description

If the rpg code contains `0` formatted as `0000 `or `00.00`, the precision of such constants, computed through `IntLiteral` or `RealLiteral`, is always 1.  

This behavior is due to the fact that:
- for the integer constant: `"0000".toLong()` is `0` and the precision is `1`
- for the real constant: `BigDecimal("00.00")` has precision `1` and scale `2`

While the first case was (more on less) predictable, the second one, honestly was not.

However, to fix this issue I have added property `precision` both for `IntLiteral` and `RealLiteral` and lightly changed the `Expression.type()` implementation.

This fix is necessary because `Expression.type()` is vital in the `EDITC` fixing which is in work in progress.

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/typesystem.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/typesystem.kt
@@ -281,7 +281,7 @@ fun Expression.type(): Type {
             NumberType(BigDecimal.valueOf(this.value).precision(), decimalDigits = 0)
         }
         is RealLiteral -> {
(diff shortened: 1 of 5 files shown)


lanarimarco:
@cosentino-smeup I have assigned you the code review because you are facing the editc issue, let me know if you need further support.

cosentino-smeup (approved):
Perfect. I am going to test it thoroughly when I will merge these changes in the branch "[bugfix/NW23001440/EDITC_problems](https://github.com/smeup/jariko/tree/bugfix/NW23001440/EDITC_problems)" because I'll have a wider number of user cases.

PR #410, opened 2024-02-01
Bugfix/nw23001440/scan with colon

davidepalladino-apuliasoft:
## Description
I tried to fix the problem by checking if the Factor 1 is literal or not. In affirmative case, is built an `StringLiteral` expression to perform the action.
In addition, has been fixed the behavior of `expression.kt` when the number of occurrences is 0.

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
@@ -1618,6 +1618,7 @@ data class ScanStmt(
         } while (index >= 0)
         if (occurrences.isEmpty()) {
             interpreter.setIndicators(this, BooleanValue.FALSE, BooleanValue.FALSE, BooleanValue.FALSE)
+            interpreter.assign(target, IntValue(0))
         } else {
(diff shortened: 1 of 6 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -1618,6 +1618,7 @@ data class ScanStmt(
         } while (index >= 0)
         if (occurrences.isEmpty()) {
             interpreter.setIndicators(this, BooleanValue.FALSE, BooleanValue.FALSE, BooleanValue.FALSE)
+            interpreter.assign(target, IntValue(0))
```
lanarimarco, line 1621: Why?

lanarimarco (changes requested):
Can you add into `SCANTEST.rpgle` an example of `SCAN` with column?
Remember that the asserts are in the `executeSCANTEST` function.

davidepalladino-apuliasoft:
> Can you add into `SCANTEST.rpgle` an example of `SCAN` with column? Remember that the asserts are in the `executeSCANTEST` function.

Sure, for next commit. I created a case where the start point is higher than length of word.

lanarimarco (approved):
Ok

PR #411, opened 2024-02-02
Bugfix/nw23001440/extname duplicated fields

cosentino-smeup:
## Description
In this branch has been fixed the wrong duplicated definition in case of 2 DS which refer to the same external file.

Following the main changes:
- fix StringType length size in case of varying value
- removed the duplicated fields in case of double external DS definition pointing to the same file
- corrected the wrong size of "entireDigits" of packed fields in the metadata file "MULANGTF.json"

Note: I modified the file MULANGTF.json manually but the wrong size of "entireDigits" of packed fields will require an additional bug fixing activity to be finally resolved. In fact, there is an external program that generates metadata json files.

Related to:
- T40_A10_P01

## Checklist:


lanarimarco (approved):
Ok

PR #413, opened 2024-02-07
Bugfix/nw23001440/in with reference

davidepalladino-apuliasoft:
## Description
I tried to fix the problem by adding to `SPLAT_INDICATOR` a case when `*IN` has access to a variable.
When I was following the evaluation of this code:
```
     C                   EVAL      *IN($IND)=*ON
```
I saw that the parser, in `misc.kt`, considered `*IN` like `IndexedTargetContext` rather than `IndicatorTargetContext`. It's wrong! `*IN`, even if it has parentheses, isn't an array later to consider `ArrayAccessExpr`. So, I decided to manipulate `RpgLexer.g4` by append to `SPLAT_INDICATOR` a case the argument of parentheses is a variable: 
```
SPLAT_INDICATOR :  
                                      ...
					| '*' [iI] [nN] '(' [§£#@$a-zA-Z]+ ')' );
```
The chose about value in the brackets is related to `ID`. A variable can contain letters and special chars like `§£#@`.
So, after this correction the parser finally consider `*IN` like `IndicatorTargetContext`. I decided to improve `TargetContext.toAst(...)`, branch `IndicatorTargetContext` of `when`, by checking the `indic` with a regex made like the latest case `SPLAT_INDICATOR` previously described. Progressively, I improved:

1. `expressions.kt` by adding `dynamicIndicatorIndex`;
2. `internal_interpreter.kt` by adding the logic of calculation of `indexExpression` if exists.
With this correction the test `T03_A10_P09-10.rpgle` passes, with other.

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/expressions.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/expressions.kt
@@ -58,7 +58,8 @@ internal fun RpgParser.IndicatorContext.toAst(conf: ToAstConfiguration = ToAstCo
         val index: Expression = (this.children[2].getChild(0) as RpgParser.ExpressionContext).toAst(conf = conf)
         return IndicatorExpr(index = index, position = toPosition(conf.considerPosition))
     } else {
-        todo(conf = conf)
+        val index = text.uppercase(Locale.getDefault()).removePrefix("*IN").toInt()
+        return IndicatorExpr(index = index, position = toPosition(conf.considerPosition))
     }
 }
 
@@ -140,6 +141,7 @@ internal fun RpgParser.IdentifierContext.toAst(conf: ToAstConfiguration = ToAstC
 
 private fun RpgParser.IdentifierContext.variableExpression(conf: ToAstConfiguration): Expression {
     return when {
+        this.text.dynamicIndicatorIndex(toPosition(conf.considerPosition)) != null -> IndicatorExpr(this.text.dynamicIndicatorIndex(toPosition(conf.considerPosition))!!, toPosition(conf.considerPosition))
         this.text.indicatorIndex() != null -> IndicatorExpr(this.text.indicatorIndex()!!, toPosition(conf.considerPosition))
         this.multipart_identifier() != null -> this.multipart_identifier().toAst(conf)
         else -> DataRefExpr(variable = ReferenceByName(this.text), position = toPosition(conf.considerPosition))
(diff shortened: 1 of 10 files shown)


lanarimarco (commented):
I did not yet see the implementation but the approach you have explained in PR comment is clear.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/expressions.kt
```diff
@@ -44,13 +44,25 @@ fun RpgParser.ExpressionContext.toAst(conf: ToAstConfiguration = ToAstConfigurat
...
         // FIXME it is rather ugly that we have to do this: we should get a different parse tree here
         this.children.size == 3 && this.children[0].text == "(" && this.children[2].text == ")"
                 && this.children[1] is RpgParser.ExpressionContext -> (this.children[1] as RpgParser.ExpressionContext).toAst(conf)
         else -> todo(conf = conf)
     }
 }
 
+internal fun RpgParser.IndicatorContext.toAst(conf: ToAstConfiguration = ToAstConfiguration()): IndicatorExpr {
+    // manage *IN(
+    if (this.children[0].text == "*IN" && this.children[1].text == "(") {
```
lanarimarco, line 57: Made it case insensitive

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
```diff
@@ -1592,10 +1592,17 @@ internal fun TargetContext.toAst(conf: ToAstConfiguration = ToAstConfiguration()
             ReferenceByName(this.getFieldName()),
             toPosition(conf.considerPosition)
         )
-        is IndicatorTargetContext -> IndicatorExpr(
-            this.indic.text.indicatorIndex()!!,
-            toPosition(conf.considerPosition)
-        )
+        is IndicatorTargetContext -> {
+            if (this.indic.text.matches(Regex("\\*IN\\([§£#@\$a-zA-Z]+\\)"))) {
```
lanarimarco, line 1596: You are assuming that `*IN(<arg>)` `<arg>` always will be a variable, but is not true.
This statement `*IN(var+1-1)` is accepted, and then you should use the same apprach that I used in the `IndicatorContext.toAst`

Add some test in `INDIC01.rpgle`

lanarimarco (changes requested):
Some changes are required.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
davidepalladino-apuliasoft: I agree with you. Focused to the task, I didn't think the case with an internal expression. In this case, I'll change `RpgLexer.g4` to consider other cases, then the logic `is IndicatorTargetContext` like you suggested.

lanarimarco (approved):
Ok very good job.

PR #417, opened 2024-02-12
Bugfix/nw23001440/dowxx operator

davidepalladino-apuliasoft:
### Main purposes
To achieve this goal, I implemented a case when the `BlockContext` is a `CsDOWxxContext`. In this case, its `toAst()` checks and creates the right `DOWxxStmt`, that take Factor 1, Factor 2, comparison operator and the body that must to be performed. Could be `DOWEQ` or `DOWGT`, etc.
The comparison is performed by `comparisonOperator.verify` function, of `comparison.kt` file, that check if the Factor 1 and 2 are `UnlimitedStringValue`, `StringValue` or other. The `compare` function has a little enhancement: now is able to process values of `UnlimitedStringValue`.
In addition, there is a test for each `DOWxx` operator, like `xx` could be `EQ`, `GT`, etc.

### Misc
- Fix: Moved the logic of comparing from `InterpreterCore.compare` to `Value` class.
- Fix: restored the logic of `Value` class.
- Fix: Removed `InterpreterCore.compare` for using `comparisonOperator.verify` like condition of `while` in `DOWxxStmt`.

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -1219,6 +1218,97 @@ data class DisplayStmt(val factor1: Expression?, val response: Expression?, over
...
 
+@Serializable
+data class DOWxxStmt(
+    val comparison: ComparisonOperator,
+    val factor1: Expression,
+    val factor2: Expression,
+    override val body: List<Statement>,
+    override val position: Position? = null
+) : Statement(position), CompositeStatement {
+    fun InterpreterCore.compare(comparison: ComparisonOperator, factor1: Expression, factor2: Expression, interpreter: InterpreterCore): Boolean {
```
lanarimarco, line 1229: You have managed a specific  comparision logic just for `DOWxx` statement, and that it means that the result of `value1 == value2 or value1 >= value2... `could be different depending on statement.
Since that you are comparing two instances of Value you should manage this logic in the `operator fun compareTo`
in the concrete implementation of Value, which is always the right place to implement this kind of logic.

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
davidepalladino-apuliasoft: So, if I understand correctly, I have to move the implementation about `.compare` into `Value`

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -1219,6 +1218,46 @@ data class DisplayStmt(val factor1: Expression?, val response: Expression?, over
...
+data class DOWxxStmt(
+    val comparison: ComparisonOperator,
+    val factor1: Expression,
+    val factor2: Expression,
+    override val body: List<Statement>,
+    override val position: Position? = null
+) : Statement(position), CompositeStatement {
+    fun InterpreterCore.compare(comparison: ComparisonOperator, factor1: Value, factor2: Value, interpreter: InterpreterCore): Boolean {
+        return when (comparison) {
+            ComparisonOperator.EQ -> factor1.compareTo(factor2) == EQUAL
```
lanarimarco, line 1231: I had asked you to implement the comparision logic in `operator fun compareTo` of the concrete class because in kotlin this signature allows to compare two instances by using the same operator you can use for scalar type and then, instead to write: `factor1.compareTo(factor2)` you can write `factor1 != factor2` which is much more expressive.
Replace `factor1.compareTo(factor2)` with` factor1 <op> factor2`

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -54,7 +54,21 @@ interface Value : Comparable<Value> {
         }
         return ConcreteArrayValue(elements, elementType)
     }
-    override operator fun compareTo(other: Value): Int = TODO("Cannot compare $this to $other")
```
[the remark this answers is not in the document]
davidepalladino-apuliasoft: Oh My Gosh! Sorry for my mistake. This was my oversight. I'll correct it.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
davidepalladino-apuliasoft: I found a solution cleaner than actual. While I was analyzing the code for searching the class for this purpose, I saw in `comparison.kt` the `ComparisonOperator?.verify` that takes as input the two factors, interpreter and charset. So, I decided to remove `InterpreterCore.compare` (I believed that was a concrete class) and using  `comparisonOperator.verify` like condition of while in `DOWxxStmt`. Of course, is loses the expressive, but analyzes the cases of `StringValue`, `UnlimitedStringValue` and other, in `comparison.kt` > `compare`. What do you think about this solution? I prepared a new commit.

lanarimarco: I did not know that function, I agree it is totally unexpressive but it makes its work.

lanarimarco (approved):
Ok

lanarimarco:
You should fix some conflicts, after that I will merge

davidepalladino-apuliasoft:
> You should fix some conflicts, after that I will merge

Yes, I'm trying but I think that I need your help. Because, after the manual merge of those `.kt` file, I have errors in several point of application at launch of `check`

lanarimarco (approved):
Ok

PR #418, opened 2024-02-12
Bugfix/nw23001440/negative variable

davidepalladino-apuliasoft:
### Main purposes
To achieve this goal, I implemented a case for `toAst` of `ExpressionContext` when the actual object is an `UnaryExpressionContext`. In this case, is called the related `toAst` that make a `NegationExpr`. This Expression make the value opposite from original.

### Misc
- Fix: Fixed description of `NEGATION.rpgle`.
- Fix: Added more cases of test for numbers.

## Checklist:


rpgJavaInterpreter-core/src/test/resources/NEGATION.rpgle
```diff
@@ -0,0 +1,32 @@
+      *
+      * WORKING WITH INDICATOR 15 AND EXPRESSIONS
```
lanarimarco, line 2: Fix description

rpgJavaInterpreter-core/src/test/resources/NEGATION.rpgle
```diff
@@ -0,0 +1,32 @@
...
+      *
+
+     D NUM             S              3  0
```
lanarimarco, line 5: Add variable `NUMD` with scale greater than zero and test it.
Tests for both `NUM` and `NUMD` variables also when negative.

Add another test where you try to negate a variable defined as string, I want to see what happens.

lanarimarco (changes requested):
Good job, but you should improve tests as suggested.

rpgJavaInterpreter-core/src/test/resources/NEGATION.rpgle
davidepalladino-apuliasoft: I implemented this case of test:
```
      *
      * NEGATION OF STRING
      *

     D STR             S             100          VARYING INZ('10')
     D STR_NEG         S             100          VARYING
     D RES             S             100          VARYING

     C                   EVAL      STR_NEG=-"STR"
     C                   EVAL      RES=%CHAR(STR_NEG)
     C* Excepted Exception
     C     RES           DSPLY
```
Rightly, the application explode in `RpgParserFacade.kt`

immagine

What is the expected result?

lanarimarco: You are not negating a variable defined as string but a string literal, and then:

- create another test program, for example: `NEGATIONERR.rpgle`
- create a test unit and use the test annotation for which the expected result is an error

davidepalladino-apuliasoft: Sorry, typo error. Now is `C                   EVAL      STR_NEG=-STR`
The result of test is:
immagine

lanarimarco (approved):
Ok

PR #419, opened 2024-02-13
Bugfix/nw23001440/Fix problem with DS 

cosentino-smeup:
## Description
In this branch we have fixed 2 problems:
1) value of DS fields: when the program retrieved the value from DS fields, did not use the real data structure container but treated the field like a single DataDefinition. This problem, lead the program to save the value in a memory area and read it from another.
 
2) DS fields with OVERLAY: when there was a definition of a DS which included both OVERLAY fields and single fields (not included in the OVERLAY) the single fields were not properly managed in the entire program.

Include a summary of the change.

Related to:
T12_A03_P06

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/SymbolTable.kt
```diff
@@ -100,8 +100,30 @@ class SymbolTable : ISymbolTable {
...
+            val containerValue = get(data.container)
+            if (data.container.isArray()) {
+                throw IllegalStateException("We do not yet handle an array container")
+            } else if (data.declaredArrayInLine != null) {
+                ProjectedArrayValue.forData(containerValue as DataStructValue, data).container.set(data, value.forType(data.type))
+                return ProjectedArrayValue.forData(containerValue as DataStructValue, data)
+            } else {
+                // Should be always a DataStructValue
+                if (containerValue is DataStructValue) {
+                    containerValue.set(data, value.forType(data.type))
```
lanarimarco, line 114: I don't understand why in `DataStructValue` you put an instance of `Value` returned by `value.forType` and after that you return the coercion of that instance.

cosentino-smeup: // step 1: the following statement set the plain value into the DS
containerValue.set(data, value.forType(data.type))

// step 2: the following statement, get plain value from the DS and coerces it to return the real type of data
return coerce(containerValue[data], data.type)

I could have returned the following value:
return value.forType(data.type)

but if something went wrong during the step 1, I could have passed the wrong value back to the caller.
In this way, I am sure that I am returning the real value from the DS.

Let me know your thoughts

lanarimarco: The matter is that we are handing the value related `DataDefinition` different by the `FieldDefinition` and I don't know what is the right way, see below:
image
However, if it works, perhaps it is not useful to ask too many questions...

cosentino-smeup: I understand your doubts, but this way to get and store values from DSs has been already used in other part of Jariko.
I think that they forgot to implement it everywhere. I just did the same thing as you can see from the image below. 

image

lanarimarco: Ok

lanarimarco (dismissed):
Ok

lanarimarco:
Before merge you shoud resolve conflicts

cosentino-smeup:
> Before merge you shoud resolve conflicts

Done

PR #420, opened 2024-02-19
Bugfix/nw23001440/like with external ds

cosentino-smeup:
## Description
In this branch has been fixed the error which occurred in case of DS field defined LIKE a field belonging to an external data definition. 

Example:
```
     DCQNCOG         E DS                  EXTNAME(CQNCOG0F) INZ
      
     DK§CF             DS
     D K_N§TINC                            LIKE(N§TINC)
```

Where N§TINC belong to CQNCOG0F

Related to:
T02_A50_p05

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/data_definitions.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/data_definitions.kt
@@ -705,7 +705,11 @@ internal fun RpgParser.Dcl_dsContext.calculateFieldInfos(knownDataDefinitions: C
     val caughtErrors = mutableListOf<Throwable>()
     val fieldsList = FieldsList(this.parm_fixed().mapNotNull {
         kotlin.runCatching {
-            it.toFieldInfo(knownDataDefinitions = knownDataDefinitions)
+            if (it.keyword().map { keyword -> keyword.keyword_like() }.firstOrNull() != null) {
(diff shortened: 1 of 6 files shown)


lanarimarco (changes requested):
Not enough tests.
You have copied `RpgParser.Parm_fixedContext.toFieldInfo` into  `RpgParser.Parm_fixedContext.toFieldInfoWithExtName` and these functions look like almost identical.
There is a lot of duplicate code, in addition, you assume that overlay and like keywords are mutually esclusive, I don't know, maybe is correct...

lanarimarco (approved):
Ok but you should start to work more thinking in koltin, for example by using optionals, scoped functions and elvis operator.
The not-null assertion operator (!!) is a dangerous way to work because can cause NullPointerException. For now is ok but since the next review I will accept "not-null assertion" and unhelpful variable declaration (scoped function) only when really needed.

These statements:
```kotlin
val isLike = this.keyword().map { keyword -> keyword.keyword_like() }.firstOrNull() != null
val keywordLike = if (isLike) this.keyword().first { it.keyword_like() != null }.keyword_like() else null
val like = if (isLike) keywordLike!!.simpleExpression().toAst(conf) as DataRefExpr else null
val varName = if (isLike) like!!.variable.name else this.name
```

could be replaced with a single statement:
```kotlin
val varName = this.keyword().firstOrNull { it.keyword_like() != null }
    ?.keyword_like()?.simpleExpression()?.toAst(conf = conf)?.let { it as DataRefExpr }?.variable?.name ?: this.name
```

where we have less variables (1 vs 4) and none "not-null assertion operator".

Remember that in a code base bugs are strictly related to: 
- Lines of codes 
- Number of variables 
- Assumptions that something is not null

PR #424, opened 2024-02-23
Bugfix/nw23001440/scan without variables

davidepalladino-apuliasoft:
## Description
In accordance with the [official documentation](https://www.ibm.com/docs/en/i/7.5?topic=codes-scan-scan-string), operator `SCAN` mightn't have a Result variable set. Before this fix, the interpreter thrown an exception because "The variable name should not blank". To fix, this problem I made optional the `target` parameter of `ScanStmt`, with several check in `execute` method.
Going in depth, by analyzing the code of case of study:
```
     C     '!'            SCAN     A35_A20A                               20
     C                   EVAL      £DBG_Str='Src1='+%CHAR(*IN20)
```
the result of `SCAN` (founded or not) is set to Indicator number 20.

## Checklist:


davidepalladino-apuliasoft:
> Try to think more in kotlin, scoped functions make the difference

Yes, I forgot the expressivity of Kotlin.

lanarimarco (approved):
Ok

PR #425, opened 2024-02-26
Bugfix/nw23001440/move indicator with variable

cosentino-smeup:
## Description
Fixed the case of IndicatorExpression with a variable expression

Related to: 
T03_A30

## Checklist:


lanarimarco (commented):
I cannot approve this PR because the issue is quite similar to #413 and then, to fix this problem, is necessary starting from that fix and try to understand why it does not cover all operation codes.

PR #426, opened 2024-02-26
Bugfix/nw23001440/operator near scala

davidepalladino-apuliasoft:
## Description
The main problem was that in an expression like `1*FOO`, `*FOO` is recognize like a variable. This isn't true! We have a multiplication between a number and a variable, where the char `*` is near the variable. To achieve the goal, is neccessary to deduce who there is before that character. The scenarios can be several, like:
- `=*VAR`
- `[a-zA-Z0-9...] *VAR` (white space between a anything and `*VAR`)
- `**VAR`
- `+*VAR`
- `-*VAR`
- `/*VAR`
- `(*VAR)`

In these cases we can assume that there is an operation with a variable that could be a pointer. So, I decided to change the `RpgLexer` from:
```
ID : ('*' {getCharPositionInLine()>7}? '*'? [a-zA-Z])?
        [§£#@%$a-zA-Z]{getCharPositionInLine()>7}? [§£#@$a-zA-Z0-9_]* ;
```
to:
```
ID : ({
            _input.LA(-1) == 32 ||
            _input.LA(-1) == '(' ||
            _input.LA(-1) == '*' ||
            _input.LA(-1) == '+' ||
            _input.LA(-1) == '/' ||
            _input.LA(-1) == '='
      }? '*' {getCharPositionInLine()>7}? '*' ? [a-zA-Z])?
      [§£#@%$a-zA-Z]{getCharPositionInLine()>7}? [§£#@$a-zA-Z0-9_]*;
```

## Checklist:


rpgJavaInterpreter-core/src/test/resources/EVALVARSNUMS.rpgle
```diff
@@ -0,0 +1,23 @@
...
+     C                   EVAL      RES=24* VAL + 6
+     C     RES           DSPLY
+
+     C                   EVAL      RES = 24* VAL + 6
+     C     RES           DSPLY
+
+     C                   EVAL      RES=24*VAL+6
+     C     RES           DSPLY
+
+     C                   EVAL      RES=24*VAL+6
```
lanarimarco, line 22: Just for curiosity, have you tried RES=24 *VAL + 6

lanarimarco (commented):
Good job, after about one hour I gave up, could you add another case in `EVALVARNUMS`?

rpgJavaInterpreter-core/src/test/resources/EVALVARSNUMS.rpgle
davidepalladino-apuliasoft: > Just for curiosity, have you tried RES=24 *VAL + 6

Yep, `*VAL` is considered like variable.

lanarimarco (approved):
Ok

PR #435, opened 2024-03-01
Bugfix/nw23001440/like static with const

lanarimarco:
The variable resolution in subrocedures through `LIKE(VAR)` syntax always went wrong when in the main at least a `D` spec as constant was declared.

## Checklist:


cosentino-smeup (approved):
Ok

PR #438, opened 2024-03-04
Bugfix/nw23001440/data reference   array with only like

davidepalladino-apuliasoft:
## Description
For the field `explicitElementType`, during construction of `FieldInfo` in `RpgParser.Parm_fixedContext.toFieldInfo`, was not considered `fields` during the search. If we have a code like this:
```
     D  A40_DS1        DS         30
     D  DS1_FL1                      2
     D  DS1_FL2                      10
     D  DS1_FL3                      15
     D  DS1_FL4                      3  0

     D  A40_DS8        DS
     D  DS8_FL1                            LIKE(DS1_FL1)
     D  DS8_FL2                            LIKE(DS1_FL2)
     D  DS8_FL3                            LIKE(DS1_FL3)
     D  DS8_FL4                            LIKE(DS1_FL4)
     D £DBG_Str        S             50          VARYING
```
the operator `LIKE` considered only `A40_DS1` and `A40_DS8`, before the fix ho this PR.

## Checklist:


cosentino-smeup (changes requested):
You could use flatMap to loop through nested arrays:
`
explicitElementType = this.calculateExplicitElementType(arraySizeDeclared, conf) ?: knownDataDefinitions.firstOrNull { it.name.equals(varName, ignoreCase = true) }?.type ?: knownDataDefinitions.flatMap { it.fields }.firstOrNull { fe -> fe.name.equals(varName, ignoreCase = true) }?.type,
`

cosentino-smeup (changes requested):
In the rpgle source files: T02_A50_P08 and T02_A50_P09

The following statement is not properly formatted: 
`
D £DBG_Str        S             150         VARYING 
`


"150" should go 1 position on the left

cosentino-smeup (approved):
Ok

PR #440, opened 2024-03-06
Bugfix/nw23001440/dow with blanks

dom-apuliasoft:
## Description

Fix an issue causing `DOWxx` statements evaluation not to produce appropriate AST when using a symbolic constant instead of a value as factor 2 causing syntax like `DOWNE *BLANKS` to fail.

Also implemented a comparison method to compare `StringValue`s with `BlanksValue`s.

Related to:
- T12_A04_P13
- T12_A04_P14

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
@@ -198,9 +198,15 @@ data class StringValue(var value: String, val varying: Boolean = false) : Abstra
         return s1.compareTo(s2)
     }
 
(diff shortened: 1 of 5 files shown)


cosentino-smeup (changes requested):
The new overload of the "compare" function, receives the following parameters that are never used: "other" and "descend". The parameter "charset" is required but never used as well. 
 
```
    fun compare(other: BlanksValue, charset: Charset?, descend: Boolean = false): Int {
        require(charset != null)
        return if (this.isBlank()) EQUAL else SMALLER
    }
```

You could consider to replace the entire function with a simple test:

```
...
is BlanksValue -> if (this.isBlank()) EQUAL else SMALLER
...
```

cosentino-smeup (approved):

PR #444, opened 2024-03-08
Bugfix/nw23001440/221 data reference with copy

davidepalladino-apuliasoft:
## Description
This solution permits to execute sources with `LIKE` and a `/copy` after the definition of variables.
The main problem is the not resolution of variable contained in `LIKE`, between its parenthesis. Considering this code (LIKEWITHCOPY1.rpgle): 
```
     D FOO             S                   LIKE(BAR)
     D BAZ             S           2560    VARYING
      /COPY QILEGEN,LIKEWITHCOPY_DEP
```
While Jariko was searching data definition, didn't find `BAR` because didn't not defined. I followed the flow of code until `getDataDefinitions` in `misc.kt`. An exception is raised during the first searching in `runCatching`. So, was decided, in accordance with Marco, to fix the problem with not another scan in `getDataDefinitions`, but a check of type in `compile_time_interpreter.kt`.

## Checklist:


lanarimarco (commented):
Maybe you have written more than necessary, try to look at [my solution](https://github.com/smeup/jariko/tree/marco/bugfix/NW23001440/221-data-reference-with-copy)

lanarimarco (approved):
Ok

PR #446, opened 2024-03-08
Bugfix/nw23001440/231 size array by variable

davidepalladino-apuliasoft:
## Description
By analyzing the code:
```
     D A30_AR4         S             20    DIM(A30_C04)
```
we have a reference that, in actual state of Jariko, isn't solved in compile time. To solve this problem, I implemented the logic to do that, in `compile_time_interpreter` where for `DataRefExpr` (in this case `A30_C04` previously defined) is called `tryToResolve` with the data definitions previously resolved. 

## Checklist:
- [x] There is a specific documentation in the `docs` directory

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
@@ -75,14 +75,10 @@ open class BaseCompileTimeInterpreter(
             is NumberOfElementsExpr -> IntValue(evaluateNumberOfElementsOf(rContext, expression.value).toLong())
             is IntLiteral -> IntValue(expression.value)
             is StringLiteral -> StringValue(expression.value)
-            is DataRefExpr -> {
-                expression.variable.tryToResolve(knownDataDefinitions)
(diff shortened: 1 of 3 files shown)


cosentino-smeup (changes requested):
1) In the following expression:
`
assertEquals(expected, "smeup/T02_A30_P04".outputOf(configuration = smeupConfig))
`
"configuration = smeupConfig" is not necessary

2) In the following RPGLE statement "150" must go 1 character on the left
`D £DBG_Str        S             150         VARYING`

3) The method "tryToResolve" returns a boolean so it's possible to simplify the following expression:
    
```
    is DataRefExpr -> {
        expression.variable.tryToResolve(knownDataDefinitions)
        if (expression.variable.referred != null) {
            return this.evaluate(rContext, (expression.variable.referred as? DataDefinition)?.initializationValue as Expression)
        }

        TODO(expression.toString())
    }

```
like this

```
    is DataRefExpr -> if (expression.variable.tryToResolve(knownDataDefinitions))
                            return this.evaluate(rContext, (expression.variable.referred as? DataDefinition)?.initializationValue as Expression)
                      else
                            TODO(expression.toString())        
```

cosentino-smeup:
Perfect! You must just resolve the conflicts.

cosentino-smeup (approved):
Ok

PR #450, opened 2024-03-12
Bugfix/nw23001440/229 data reference inline do

davidepalladino-apuliasoft:
## Description
To fix this problem, was necessary to improve `findInStatementDataDefinitions` in `resolution.kt`. The main view is that a statement could have a `dataDefinition` (for inline variable) but also a `body` with nested statements with other `dataDefinition`. With this fix is resolved task `229`.

But task `230` is particular. Considering this code:
```
     D A20_D1          S              3  0 INZ(32)
     D A20_D2          S              3  1 INZ(0.6)

     C     A20_D1        DIV       A20_D2        A20_D7           11 2
     C                   MVR                     A20_D8            5 3
```
here we have a `MVR` with a data declared inline. To fix this problem, `MVR` must be a statement, specifically a `StatementThatCanDefineData`. For `DivStmt` is not necessary to know only the `result` like a variable, but also like a `dataDefinition` if is declared inline like other cases. For this reason, `mvrResult` is replaced with `mvrStatement`. With this fix is resolved task `230`

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
@@ -717,19 +717,11 @@ open class InternalInterpreter(
         // calculation of rest
         // NB. rest based on type of quotient
         if (statement.mvrStatement != null) {
-            val restType = when {
-                statement.mvrStatement.dataDefinition != null -> statement.mvrStatement.dataDefinition.type
-                statement.mvrStatement.target != null -> statement.mvrStatement.target.type()
-                else -> null
-            }
+            val restType = statement.mvrStatement.target?.type()
             require(restType is NumberType)
             val truncatedQuotient: BigDecimal = quotient.setScale(type.decimalDigits, RoundingMode.DOWN)
-            // rest = divident - (truncatedQuotient * divisor)
             val rest: BigDecimal = dividend.subtract(truncatedQuotient.multiply(divisor))
-            when {
-                statement.mvrStatement.dataDefinition != null -> assign(statement.mvrStatement.dataDefinition, DecimalValue(rest.setScale(restType.decimalDigits, RoundingMode.DOWN)))
-                statement.mvrStatement.target != null -> assign(statement.mvrStatement.target, DecimalValue(rest.setScale(restType.decimalDigits, RoundingMode.DOWN)))
-            }
+            assign(statement.mvrStatement.target, DecimalValue(rest.setScale(restType.decimalDigits, RoundingMode.DOWN)))
         }
         return if (statement.halfAdjust) {
             DecimalValue(quotient.setScale(type.decimalDigits, RoundingMode.HALF_UP))
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
@@ -716,13 +716,20 @@ open class InternalInterpreter(
         require(type is NumberType)
         // calculation of rest
         // NB. rest based on type of quotient
-        if (statement.mvrTarget != null) {
-            val restType = statement.mvrTarget.type()
+        if (statement.mvrStatement != null) {
+            val restType = when {
+                statement.mvrStatement.dataDefinition != null -> statement.mvrStatement.dataDefinition.type
+                statement.mvrStatement.target != null -> statement.mvrStatement.target.type()
+                else -> null
+            }
             require(restType is NumberType)
             val truncatedQuotient: BigDecimal = quotient.setScale(type.decimalDigits, RoundingMode.DOWN)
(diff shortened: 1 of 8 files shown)


cosentino-smeup (changes requested):
1) I had some difficulties to understand which commit was related to the task 229 and which to 230. Maybe next time could be simpler to split the tasks in 2 separated branches even if there is a common problem that connect them.

2) Good idea transforming the MVR expression in a statement, which by the way, was already foreseen in the parser, but the property StatementThatCanDefineData could have been used differently. 

3) In the source file internal_interpreter.kt, function "div":   

```
        if (statement.mvrStatement != null) {
            val restType = when {
                statement.mvrStatement.dataDefinition != null -> statement.mvrStatement.dataDefinition.type
                statement.mvrStatement.target != null -> statement.mvrStatement.target.type()
                else -> null
            }
            require(restType is NumberType)
            val truncatedQuotient: BigDecimal = quotient.setScale(type.decimalDigits, RoundingMode.DOWN)
            // rest = divident - (truncatedQuotient * divisor)
            val rest: BigDecimal = dividend.subtract(truncatedQuotient.multiply(divisor))
            when {
                statement.mvrStatement.dataDefinition != null -> assign(statement.mvrStatement.dataDefinition, DecimalValue(rest.setScale(restType.decimalDigits, RoundingMode.DOWN)))
                statement.mvrStatement.target != null -> assign(statement.mvrStatement.target, DecimalValue(rest.setScale(restType.decimalDigits, RoundingMode.DOWN)))
            }
        }
```
        
if "statement.mvrStatement != null" then "statement.mvrStatement.target" will never be null because it's mandatory. Consider to make these changes to semplify the code:

```
	if (statement.mvrStatement != null) {
            val restType = statement.mvrStatement.target?.type()
            require(restType is NumberType)
            val truncatedQuotient: BigDecimal = quotient.setScale(type.decimalDigits, RoundingMode.DOWN)
            val rest: BigDecimal = dividend.subtract(truncatedQuotient.multiply(divisor))
            assign(statement.mvrStatement.target, DecimalValue(rest.setScale(restType.decimalDigits, RoundingMode.DOWN)))
        }

```

4) In the source file resolution.kt

- function "findInStatementDataDefinitions":

I believe the following comment could now be removed:
`// TODO could they be also annidated?`

- function "scanInDephStatementDataDefinition":

There is something wrong with this loop: 

```
 node.body.forEach {
                if (it is StatementThatCanDefineData) {
                    list.addAll(it.dataDefinition())
                } else if (it.parent != null && it.parent is StatementThatCanDefineData) {
                    list.addAll((it.parent as StatementThatCanDefineData).dataDefinition())
                } else {
                    list.addAll(scanInDephStatementDataDefinition(it))
                }
            }

```
If I put 2 statements inside the DO loop

image

the program is gonna add the same variable two times

image

In general, I think that this function should perform the following operations:
- when it is a simple statement and there is a data definition, it should be added to the list
- when it is a "CompositeStatement" the program should:
     - add the data definition on the block (if StatementThatCanDefineData)
     - loop through the statements of the block an invoke itself recursively a passing the current statement as parameter 

I don't like the following test because it is not generic:

```
          if (node is DivStmt) {
                node.mvrStatement?.let { list.addAll(it.dataDefinition()) }
            }
```
Isn't it possible to take advantage of the property StatementThatCanDefineData inside of MvrStmt?

davidepalladino-apuliasoft:
> 1. I had some difficulties to understand which commit was related to the task 229 and which to 230. Maybe next time could be simpler to split the tasks in 2 separated branches even if there is a common problem that connect them.

Yes, I worked on same branch for two task because the solution of one is related to other one. In other case I should close the task "229" to continue for the "230" that request the solution of first; can be time consuming.

davidepalladino-apuliasoft:
> 2\. Good idea transforming the MVR expression in a statement, which by the way, was already foreseen in the parser, but the property StatementThatCanDefineData could have been used differently.

I didn't understand what do you mean. I used `StatementThatCanDefineData` because could be a situation where the result could be defined inline instead previously.

davidepalladino-apuliasoft:
> if "statement.mvrStatement != null" then "statement.mvrStatement.target" will never be null because it's mandatory. 

Instead it is exactly the opposite: when the result wasn't defined previously but inline.

davidepalladino-apuliasoft:
> the program is gonna add the same variable two times

For this, I should add something like `addIfDoesntExist`

davidepalladino-apuliasoft:
> Isn't it possible to take advantage of the property StatementThatCanDefineData inside of MvrStmt?

Unfortunately I think isn't possible because when is called a `div`, should there be the resolution of its MVR that could have a data definition inline. If you want, we can try together in a Meet.

cosentino-smeup:
> > if "statement.mvrStatement != null" then "statement.mvrStatement.target" will never be null because it's mandatory.
> 
> Instead it is exactly the opposite: when the result wasn't defined previously but inline.

What I meant is that with RPG you can't define an MVR operation code without the the target (result factor)

image

That means, that it doesn't matter if you define the variable of the target in line or in the data definitions statement (D) or in line with another statement, the target will be always defined. I tested the different cases it looks like this. Please let me know if there is a case where you need the  "statement.mvrStatement.dataDefinition" because "statement.mvrStatement.target" is null. Maybe I missed it.

cosentino-smeup (approved):
ok

PR #454, opened 2024-03-13
Bugfix/nw23001440/238 data reference to procedure

davidepalladino-apuliasoft:
## Description
In this context, a procedure could be called by its name without parenthesis. Jariko, in this case, recognizes the name like identifier instead a function. So, in `expressions` the is a check if the name is a procedure or not.

## Checklist:


cosentino-smeup (approved):
Ok. I have also tested the case of a procedure defined into an external RPGLE source file imported by /COPY and it works.

PR #456, opened 2024-03-13
Bugfix/nw23001440/scan evaluation

dom-apuliasoft:
## Description

Extend ScanStmt in order to accept array indexing and expressions after colons.

Related to:
- T10_A35_P08
- T10_A35_P10

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
```diff
@@ -1316,20 +1316,30 @@ internal fun CsDELETEContext.toAst(conf: ToAstConfiguration): Statement {
 
 internal fun CsSCANContext.toAst(conf: ToAstConfiguration = ToAstConfiguration()): ScanStmt {
     val position = toPosition(conf.considerPosition)
+
     val (compareExpression, compareLength) = this.factor1Context().toIndexedExpression(conf)
```
lanarimarco, line 1320: I would have attempted to refactor the extension function private fun FactorContext.toIndexedExpression rather than reimplementing its logic. I've noticed that it is utilized in other operation codes (try finding its usage), and perhaps we could address the same issue across those operation codes in one go. However, I'm unsure if these operation codes are affected by the same issue. You should verify this. If so, consider refactoring FactorContext.toIndexedExpression to return a Pair<Expression, Expression?>.

dom-apuliasoft: That is a fair point. I split the implementation because I was not sure the same logic would be legal for `factor1`s. However I'll commit to a further investigation soon (and try to refactor).

lanarimarco (changes requested):
Move test in `MULANGT10BaseCodopTest`

PR #472, opened 2024-03-21
Bugfix/nw23001440/251 data definition parm ds

davidepalladino-apuliasoft:
## Description
The problem was that Jariko was tightened between two different data types but with same size. By analyzing this code:
```
     D A60_D3          DS
     D  A60_DA1                       5
     D  A60_DA2                      45

[...]
     C                   PARM                    A60_D3           50
[...]
```
we have a definition in `D` as Data Structure, and an inline definition (with the same name and same size) as String. In RPG this is legal when the size (in memory allocation too) is equal. So, I fixed the problem firstly in `parsetreetoast/misc.kt`, accepting a case when have the same size and when the types are Data Structure and String. Then in `interpreter/ExpressionEvaluation.kt` for `EVAL` evaluation.
In accord with Nicolò, has been created a test with Boolean and String types, where the logic and fix are same.

In these cases, wins that defined in `D`.

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
```diff
@@ -2037,14 +2039,30 @@ internal fun <T : AbstractDataDefinition> List<T>.removeDuplicatedDataDefinition
...
+            it.require(it.matchType(dataDefinition)) {
                 "Incongruous definitions of ${it.name}: ${dataDefinition.type} vs ${it.type}"
             }
             false
         }
     }
 }
 
+internal fun AbstractDataDefinition.matchType(dataDefinition: AbstractDataDefinition): Boolean {
+    if (dataDefinition.type == this.type) {
```
lanarimarco, line 2051: I woud have implemented it more or less like this, can you try and let me know what you think:
```kotlin
private fun AbstractDataDefinition.matchType(dataDefinition: AbstractDataDefinition): Boolean {
    return when {
        dataDefinition.type == this.type -> true
        dataDefinition.elementSize() == this.elementSize() -> {
            dataDefinition.type is StringType && this.type is DataStructureType ||
            dataDefinition.type is DataStructureType && this.type is StringType ||
            dataDefinition.type is BooleanType && this.type is StringType ||
            dataDefinition.type is StringType && this.type is BooleanType
        }
        else -> false
    }
}
```

davidepalladino-apuliasoft: Works fine

lanarimarco (approved):
OK

PR #475, opened 2024-03-22
Bugfix/nw23001440/252 data definition inline video

davidepalladino-apuliasoft:
## Description
For `NumericType`, `rpgType` attribute is facoltative. So, operator "==" is tightened. So, I created an override of `Equals` where the check between `rpgType` is possible only both aren't empty.

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/typesystem.kt
```diff
@@ -193,7 +193,7 @@ infix fun Long.log(base: Int): Double {
 }
 
 @Serializable
-data class NumberType(val entireDigits: Int, val decimalDigits: Int, val rpgType: String? = "") : Type() {
+data class `NumberType`(val entireDigits: Int, val decimalDigits: Int, val rpgType: String? = "") : Type() {
```
lanarimarco, line 196: Why backtick?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/typesystem.kt
```diff
@@ -242,6 +242,19 @@ data class NumberType(val entireDigits: Int, val decimalDigits: Int, val rpgType
             return false
         }
     }
+
+    override fun equals(other: Any?): Boolean {
+        return if (other is NumberType) {
+            val resultDigits = this.entireDigits == other.entireDigits && this.decimalDigits == other.decimalDigits
+            if (rpgType?.isNotBlank()!! && other.rpgType?.isNotEmpty()!!) {
```
lanarimarco, line 249: I don't understand why in order to test `rpgType`. `this.rpgType` must be not blank while `other.rpgType` must be not empty.

lanarimarco (approved):

PR #479, opened 2024-03-27
Bugfix/nw23001440/261 percent len near eval

davidepalladino-apuliasoft:
## Description
In accord to [official documentation](https://www.ibm.com/docs/en/i/7.5?topic=functions-len-get-set-length), with `%LEN` is possible to set a size of a data definition, too; this means that is executed at runtime. 
This requires some change of implementation, by changing access of certain attributes, of several classes, from `val` to `var`. Also the `RpaParser.g4` was changed by adding `bif_len` as target. Then, has been implemented the logic of set with `%LEN`.

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/interpretation_utils.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/interpretation_utils.kt
@@ -66,7 +66,7 @@ fun ActivationGroupType.assignedName(current: RpgProgram, caller: RpgProgram?): 
     }
 }
 
-fun DataDefinition.resizeStringSize(newSize: Int): Unit {
+fun DataDefinition.resizeStringSize(newSize: Int) {
     require(this.initializationValue is StringLiteral)
     require(this.defaultValue is StringValue)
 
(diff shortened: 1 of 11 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/interpretation_utils.kt
```diff
@@ -65,3 +65,20 @@ fun ActivationGroupType.assignedName(current: RpgProgram, caller: RpgProgram?):
         else -> error("$this ActivationGroupType is not yet handled")
     }
 }
+
+fun DataDefinition.resizeStringSize(newSize: Int) {
```
lanarimarco, line 69: If we don't have to expose Jariko features, it is preferable to use the `internal` modifier, in addition I would like a little bit of documentation for both public and internal functions too, mostly for helper and utilty functions.
Take in account of this for the future.

davidepalladino-apuliasoft: I added a bit of documentation. Tell me if is ok.

lanarimarco: You should explain the logic, for instance: the method transforms the type from varying to not varying and if the target size is shorter the content will be truncated...

davidepalladino-apuliasoft: Done

lanarimarco (approved):

PR #483, opened 2024-04-03
Feature/improve api

davidepalladino-apuliasoft:
## Description
This work had the purpose to check and improve the logic of `/API` directive. Was prepared a documentation of execution's flow about the utilization, in particular for the twice data definition between main source code and that "imported" with directive.
Moreover. was tested both the procedures and the subroutines, with twice definition like the previous case. For the procedures there was a little rework: firstly by getting those in `api.kt`, like for data definitions; secondary, about the method for understanding if a procedure is fake or not. To be more exhaustive, we consider this example:

```
     DPROC             PR
...
      *
       /API APIPROCS
      *
...
     C                   CALLP     PROC(a:b:c)
...
```
This main source doesn't have the definition `PROC`. So, this `PR` is considered like "fake"; maybe injected from test, I don't know. But in reality is defined into `APIPROCS` file. The logic of construction of procedures (in `cu_components.kt`) doesn't indicate if a procedure is fake or not. So, I added a method to indicate it, necessary for `includeProceduresWithoutDuplicates` to construct the new procedures list by considering both main source and related by directive.
Moreover, there is an important improvement about the method `getFakeProcedures(...)` with a fixing, refactoring and improving for creation of fake procedures' list. The previous implementation considered only one procedure prototype. Wasn't correct. In addition, the filtering with the concrete implementation happened after the creation of the objects related to prototype procedure; was an unnecessary process with several unnecessary memory allocation.

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/cu_components.kt
```diff
@@ -42,6 +42,7 @@ data class CompilationUnit(
     // necessary to avoid that during data references resolving, are considered expression declared within procedures as well.
     @property:Link val procedures: List<CompilationUnit>? = null,
     val procedureName: String? = null,
+    val isFakeProcedure: Boolean = false,
```
lanarimarco, line 45: I am not sure if this new property is really necessary, but maybe I am wrong.
If you aim to determine whether 'CompilationUnit' is a procedure or not, I think you may test 'CompilationUnit.procedureName', which is not null only if the 'CompilationUnit' is related to a procedure.

lanarimarco (commented):
I have doubts regarding `CompilationUnit.isFakeProcedure` property.
You should add also, both for subroutine and procedure two test about, how to write a program that uses correctly the `/API` directive to include subroutines and procedures.
The current tests, which are necessary because we always have to try the features at the worst case, are perfect, but I want to see also how to write a good snippet of program...

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/cu_components.kt
davidepalladino-apuliasoft: Is necessary because there isn't any flag that indicates if a procedure is "real" or "fake". With a flag like that, I can resolve a procedure declared like fake but with its implementation imported after and by directive.
I can try to make that attribute optional, with a value if the `CompilationUnit` refers to a procedure.

davidepalladino-apuliasoft:
> I have doubts regarding `CompilationUnit.isFakeProcedure` property. You should add also, both for subroutine and procedure two test about, how to write a program that uses correctly the `/API` directive to include subroutines and procedures. The current tests, which are necessary because we always have to try the features at the worst case, are perfect, but I want to see also how to write a good snippet of program...

Yes, I can create a program with procedure, subroutine and data definition both in main program and in imported by directive.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
```diff
@@ -251,7 +251,6 @@ fun RContext.toAst(conf: ToAstConfiguration = ToAstConfiguration(), source: Stri
     val fakeProcedures = getFakeProcedures(rContext = this,
```
lanarimarco, line 251: Considered your refactory and the fact that `fakeProcedures`, is a list of `CompilationUnit` with no statement, don't you think that the right name, rather than `getFakeProcedures` should be `getProcerurePrototypes?`

lanarimarco (commented):
This issue was born to test `/API` directive. I remember a unit test `MU711001`, can you try it if it works properly?

lanarimarco (approved):
ok

PR #487, opened 2024-04-05
Bugfix/nw23001440/160 like defined inline

davidepalladino-apuliasoft:
## Description
This work fixes the utilization of `DEFINE` operator when is used into a subroutine. The main problem was the resolution of `InStatementDataDefinition` that happened from main statements and not from the subroutine.
In addition, were added the mocked `.rplge` files used by atomic tests.

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/test/resources/QILEGEN/MULANG_D_C.rpgle
+++ b/rpgJavaInterpreter-core/src/test/resources/QILEGEN/MULANG_D_C.rpgle
@@ -1,0 +1,9 @@
+      * This is a dummy and mocked implementation.
+     C     £DBG          BEGSR
+
+     C     *BLANK        IFNE       £DBG_Str
+     C     £DBG_Str      DSPLY
+     C                   EVAL       £DBG_Str = *BLANKS
+     C                   ENDIF
+
+     C                   ENDSR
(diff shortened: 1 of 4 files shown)


rpgJavaInterpreter-core/src/test/resources/QILEGEN/MULANG_D_C.rpgle
```diff
@@ -0,0 +1,9 @@
+      * This is a dummy and mocked implementation.
+     C     £DBG          BEGSR
+
+     C     *BLANK        IFNE       £DBG_Str
```
lanarimarco, line 4: Why do you test whether is not blank?

davidepalladino-apuliasoft: I was waiting for this question. By considering the main of atomic test:
```
     C                   EVAL      £DBG_Pgm = 'MU025002'
     C                   EVAL      £DBG_Sez = 'A50'
     C                   EVAL      £DBG_Fun = '*INZ'
     C                   EXSR      £DBG
     C                   EXSR      SEZ_A50
     C                   EXSR      £DBG
     C                   EVAL      £DBG_Fun = '*END'
     C                   EXSR      £DBG
     C                   SETON   
```
we have the `£DBG` that is executed three times. At the 1st time , the `£DBG_Str` is blank; at the 2nd time we have the expected result;  at the 3rd time we have the same result of 2nd time. So, we have:
```
""
"A50_A3( ) A50_A4( )"
"A50_A3( ) A50_A4( )"
```

lanarimarco: Ok

lanarimarco (approved):

PR #489, opened 2024-04-09
Bugfix/nw23001440/266 ds with like overlay

davidepalladino-apuliasoft:
## Description
The main problem was the resolution of a field of DS that uses `LIKE` operator of a previous data definition. The implementation of `getDataDefinitions` scans firstly the `dcl_ds` structure and then `dspec`. This causes the not resolution about a data definition, might be resolved after. So, @lanarimarco has implemented a fix by searching the unknown data definition during the construction of field, by injecting `InjectableCompileTimeInterpreter` and by calling `evaluateTypeOf` that resolves the `explicitElementType`.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,6 @@
 ![jariko Logo](/images/jariko_small.png)  
 
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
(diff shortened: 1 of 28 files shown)


davidepalladino-apuliasoft:
This branch is a variant, a previously work: https://github.com/smeup/jariko/tree/bugfix/NW23001440/266-ds-with-like-overlay-variant

lanarimarco (changes requested):
Revert changes in `errors.kt` and `misc.kt` and apply changes as suggested via chat.

lanarimarco (approved):
Ok

PR #498, opened 2024-04-18
Bugfix/nw23001440/and near comment

davidepalladino-apuliasoft:
## Description
This work, done with @dom-apuliasoft, fixes the problem where a long condition is followed by a comment, without any space. While we were studying the parse tree, we saw `ANDCOSTANTE` like `identifier` and not under `<beginif>`.

```
...
<statement>
  <free>
    <baseExpression>
      <expression>
        <identifier>
          ANDCOSTANTE
        </identifier>
      </expression>
    </baseExpression>
    &lt;missing FREE_SEMI&gt;
  </free>
</statement>
...
```
So, to solve the problem we added another condition for `ID` in `RpgLexer.g4`

## Checklist:


lanarimarco (approved):
Ok

PR #499, opened 2024-04-19
Bugfix/nw23001440/bif elem into occurs

davidepalladino-apuliasoft:
## Description
An `OCCURS` might have an expression to evaluate. So, I applied a fix into `RpgParser` by adding a case where the argument is an expression, in addition to `numeric_constant`. Then, I re-built the logic of evaluation in `data_definitions` by considering both numeric and expression.

## Checklist:


lanarimarco (approved):
Ok

PR #505, opened 2024-04-24
Handle IF DEFINED compiler directives with special parameters as unsupported

foresti-smeup:
## Description

Handle IF DEFINED compiler directives with special parameters as unsupported, throwing a CompilerDirectiveException.

Related to the 'Unexpected ENDIF directive' error, observed in all programs that include £CRI as a /COPY, as well as in other instances.

## Checklist:


lanarimarco (approved):
Ok

PR #508, opened 2024-05-02
Perf/logging regression

dom-apuliasoft:
## Description

Mitigate performance regression after logging system refactor

Related to #495 

## Checklist:
- [x] There are tests regarding this feature
- [x] The code follows the Kotlin conventions (run `./gradlew ktlintCheck`)
- [x] The code passes all tests (run `./gradlew check`)
- [ ] There is a specific documentation in the `docs` directory

PR #495, opened 2024-04-16
Refactor/logging

dom-apuliasoft:
## Description

Complete remake of the logging system

## Checklist:
- [x] There are tests regarding this feature
- [x] The code follows the Kotlin conventions (run `./gradlew ktlintCheck`)
- [x] The code passes all tests (run `./gradlew check`)
- [x] There is a specific documentation in the `docs` directory

PR #510, opened 2024-05-02
Feature/improve analytics logs

dom-apuliasoft:
## Description

Improve analytics channel output.

Related to #495, #508

## Checklist:


rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/smeup/PerfLoggingExample.kt
```diff
@@ -16,4 +16,36 @@ class PerfLoggingExample : MULANGTTest() {
...
+        val si = JavaSystemInterface().apply {
+            loggingConfiguration = consoleLoggingConfiguration(LogChannel.ANALYTICS)
+        }
+        executePgm(programName = "smeup/T10_A60_P03", systemInterface = si)
+    }
+
+    @Test @Category(PerformanceTest::class)
+    fun testT40_A20_P17() {
+        val si = JavaSystemInterface().apply {
+            loggingConfiguration = consoleLoggingConfiguration(LogChannel.ANALYTICS)
```
lanarimarco, line 39: In this test, there is no trace of the symbol table's get action, despite the DO statement continuously accessing the symbol table to evaluate the exit condition. If we want to be sure not to lose anything, we have to implement this logic in the SymbolTable.get and SymbolTable.set methods. You can do this in the concrete implementation com.smeup.rpgparser.interpreter.SymbolTable because com.smeup.rpgparser.experimental.SymbolTable is deprecated and was nothing more than an experiment.

lanarimarco (changes requested):
Review the `PERF` and `ANALYTICS` documentation. It is important that we provide information about what those numbers represent. 
Fix the failing test.

lanarimarco (approved):
Ok

PR #511, opened 2024-05-03
Bugfix/nw23001440/loop resolution

davidepalladino-apuliasoft:
## Description
The main problem was the resolution of data structure's params, that happened by calling `toAst` of the entire data structure. This wrong solution caused a loop of calls of the same data structure that, for a field, utilizes the `LIKE` to a field of data structure not yet processed.
To avoid this problem, the new implementation calls the `toAst` of single field.
Furthermore, there is a fix for the resolution of `ArrayType` with a value that is not an `ArrayValue` but a `IntValue`.
Finaly, there are several tests; in particular, a _smoke test_ to check the creation of fields for a DataStructure that utilizes `OVERLAY`.

## Checklist:


lanarimarco (approved):
Ok

PR #512, opened 2024-05-07
Bugfix/nw23001440/like referencing PARM inline definition

dom-apuliasoft:
## Description

Allow D-spec definition using LIKE keyword referencing an inline data definition provided by a PARM in PlistStmt.

Related to:
- MU106011

## Checklist:


lanarimarco (commented):
Good  because you made just few changes.
I would ask you a little fix in PR description because in my opinion is too generic.
In this PR you didn't fix all cases where a `D SPEC` is defined in whatever _**statement that can define data**_, but only when _**that _statement that can define data**_ is a `PARM` operation code.
You should fix the comment.

lanarimarco (approved):
Ok

PR #280, opened 2021-11-10
First official jariko release

lanarimarco:
## Description

First official Jariko release

## Checklist:
- [X] There are tests regarding this feature
- [X] The code follows the Kotlin conventions (run `./gradlew ktlintCheck`)
- [X] The code passes all tests (run `./gradlew check`)
- [X] There is a specific documentation in the `docs` directory

PR #517, opened 2024-05-14
Bugfix/LS24002645/LIKE to file field with utilization of DS

davidepalladino-apuliasoft:
## Description
To achieve the goal, I took from [#278](https://github.com/smeup/jariko/tree/bugfix/NW23001440/278-clear-and-data-reference) and [#280](https://github.com/smeup/jariko/tree/bugfix/NW23001440/280-data-reference-to-ds) branches, with Cherry Pick`, several implementations for:
 - resolution of data reference in DS that uses a `LIKE` to field of file (from 280);
 - connection to DB, with creation and populating of table (from 278).

So, the test was adapted by using `MULANGTL` instead `AUTOOG1L`, to avoid special char for columns names.

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
```diff
@@ -154,6 +154,12 @@ open class BaseCompileTimeInterpreter(
     private fun findSize(statements: List<RpgParser.StatementContext>, declName: String, conf: ToAstConfiguration, innerBlock: Boolean = true): Int? {
         statements.forEach {
             when {
+                it.fspec_fixed() != null -> {
+                    val size = it.fspec_fixed().runParserRuleContext(conf) { context ->
+                        kotlin.runCatching { context.toAst(conf).let { dataDefinition -> dataDefinition to dataDefinition.toDataDefinitions() } }.getOrNull()
```
lanarimarco, line 159: why did you convert the instance of `FileDefinition` returned by `rpgParser.Fspec_fixedContext.toAst() `in a `Pair` when you use only e second element of this `Pair?`
It was simpler:
```kt
                it.fspec_fixed() != null -> {
                    val size = it.fspec_fixed().runParserRuleContext(conf) { context ->
                        kotlin.runCatching { context.toAst(conf).let { fileDefinition -> fileDefinition.toDataDefinitions() } }.getOrNull()
                    }?.find { dataDefinition ->  dataDefinition.name.equals(declName, ignoreCase = true) }?.elementSize()
                    if (size != null) return size
                }
```
where I have also replaced the lambda argument `it` with a more espressive name.
Remember the names are crucial for readibility and maintenance

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
```diff
@@ -234,6 +240,12 @@ open class BaseCompileTimeInterpreter(
         statements
             .forEach { it ->
                 when {
+                    it.fspec_fixed() != null -> {
```
lanarimarco, line 243: see above

lanarimarco (changes requested):
I have some questions

lanarimarco (approved):
Ok

PR #524, opened 2024-05-24
Bugfix/ls24002766/douxx with indicators

davidepalladino-apuliasoft:
## Description
This work resolves a problem where a factor could be a constant (like `IN`) instead a content.
In addition, I built a new `executePgmCallBackTest` that checks both line and message expected.

## Checklist:


rpgJavaInterpreter-core/src/test/resources/ERROR23.rpgle
```diff
@@ -0,0 +1,16 @@
+     V* ==============================================================
+     D* 09/05/24
+     D* Purpose: Must fire the following errors
+     D* line 9 - Factor 1 can't be null
+     D* line 14 - Factor 2 can't be null
+     V* ==============================================================
+
+     C                   SETOFF                                           55
+     C     *IN55         DOUEQ
```
lanarimarco, line 9: Ask rpgle developer if `*IN55` or `*ON` have no default value, because I am not so sure it is right that in this case the program must throw an error

lanarimarco (commented):
I have some doubts about `ERROR23.rpgle`

rpgJavaInterpreter-core/src/test/resources/ERROR23.rpgle
davidepalladino-apuliasoft: I'll ask!

davidepalladino-apuliasoft: I run it on AS400 and this is the result:

Schermata del 2024-05-28 08-07-31

lanarimarco: Ok

lanarimarco (approved):
Ok

PR #531, opened 2024-05-30
Bugfix/ls24002827/ds extname with additional field

davidepalladino-apuliasoft:
## Description
This work resolves the construction of a DS which utilizes `EXTNAME` to get its fields; then to add another field. Before this fix (after a little fix for `Parm_fixedContext.toFieldInfo` related the `LIKE`) the execution raises an `ErrorEvent` for twice definition. 

Schermata del 2024-05-29 15-35-40

Why this?  `List<StatementContext?>.getDataDefinition` firstly scans the DS by ignoring `EXTNAME`; in this case builds `A40DS1` with its field. Then, re-scans the DS with the `EXTNAME`. The problem might be resolved by merging the two DS into one but, in my opinion, is incorrect. In this case is necessary in one step and in order:
1. to check the existence of `EXTNAME` and to get its field;
2. to get DS' fields;
3. to calculate sequences and offsets.

Finally, the logic for `LIKEDS` has been moved into one and dedicated `toAst` for DS.

## Checklist:


lanarimarco (changes requested):
I don't understand because it was enough fixing  `RpgParser.Dcl_dsContext.toAstWithExtName` 

Try to fix issue by changing `toAstWithExtName` because is the more natural place., at least in my mind.

lanarimarco (approved):
Ok

lanarimarco:
There are some conflicts to fix.

lanarimarco (approved):
Ok

PR #539, opened 2024-06-04
Feature/ls24002887/bif lookupxx

dom-apuliasoft:
## Description

Implement missing `%LOOKUPxx` BIFs. Each implementation should now correctly evaluate edge cases for:
- multiple occurrences of the same value 
- array marked as sequenced

Related to:
- LS24002887

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -1071,6 +1072,14 @@ data class DataStructValue(var value: String, private val optionalExternalLen: I
 fun Int.asValue() = IntValue(this.toLong())
 fun Boolean.asValue() = BooleanValue(this)
 
+infix fun Value.distanceFrom(other: Value): Long {
+    val a = this.asInt()
+    val b = other.asInt()
+    return (b - a).value
+}
+
+infix fun Value.absoluteDistanceFrom(other: Value): Long = abs(this distanceFrom other)
```
lanarimarco, line 1081: I think it is never used, just like `distanceFrom`. Do you believe it could be useful in the future?

lanarimarco (commented):
Very elegant way to implement things. Just a question about two new infix functions in the module `values.kt` that seem to be never used.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
dom-apuliasoft: You are right. I think it is a leftover of one of the first draft implementation of the binary search strategy but as of now they are unused. I removed them in the latest commit as they are so easy to implement that I think rewriting them if needed in future outweights keeping them unreferenced in the code for no concrete present reason.

lanarimarco (approved):
Ok

PR #541, opened 2024-06-07
Bugfix/ls24002930/define instatement defs

dom-apuliasoft:
## Description

Add support for DEFINE Statement based on instatement data definitions.

Related to:
- LS24002930

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -1188,7 +1188,10 @@ data class DefineStmt(
         val containingCU = this.ancestor(CompilationUnit::class.java)
             ?: return emptyList()
 
+        // Search standalone 'D spec' or InStatement definition
```
lanarimarco, line 1191: I would have used the computed property `CompilationUnit.allDataDefinitions`, because it contains also `inStatementsDataDefinitions`. Could you try?

PR #543, opened 2024-06-10
Bugfix/ls24002936/non blocking comptime interpreter

dom-apuliasoft:
## Description

Made CompileTimeInterpreter `findSize` and `findType` methods non-blocking on error when looking for declaration in statements. This was necessary to prevent missing declarations after statements with errors due to a premature control-flow exit in a similar fashion to what already happens in the standard `getDataDefinitions` method.

Related to # (issue)

## Checklist:


rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00214.rpgle
```diff
@@ -0,0 +1,13 @@
...
+     D* 07/06/24  nn.mm author
+     D* D spec defined LIKE PARM is not resolved when:
+     D* DEFINE statement throws an error: Error while creating data definition
+     V* ==============================================================
+     D MSG             S             10
+
+     D OLDIPG          S                   LIKE(§CRIPG)
+     C     OLDIPG        DSPLY
+     C     DTAARA        DEFINE    *LDA          £UDLDA
+     C*                   IN        £UDLDA
```
lanarimarco, line 11: Why is `IN` stat commented?

lanarimarco (commented):
The test unit related to how Jariko handles errors must be implemented like this:
- Create `ERROR<nn>.rpgle` in `rpgJavaInterpreter-core/src/test/resources/`.
- In the prologue of the file, describe the expected error in terms of which lines you expect Jariko to fire an `ErrorEvent`.
- Implement `executeERROR<nn>SourceLineTest` (copy and paste `executeERROR<mm>SourceLineTest`, replacing `<mm>` with `<nn>`).
- Implement `executeERROR<nn>CallBackTest` (copy and paste `executeERROR<mm>CallBackTest`, replacing `<mm>` with `<nn>`).

Some good examples are the tests written by @davidepalladino-apuliasoft with `<mm> = 23`.

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00214.rpgle
dom-apuliasoft: It's commented because the `IN` statement is not implemented yet, causing `gradlew check` to fail prematurely. As it is not necessary to replicate the issue I commented it. We can also consider removing it from the test or adding a mock implementation of the `IN` statement.

lanarimarco: **gradlew check failure**
You could bypass this error by adding your program name in `testing_utils.kt` `(whiteListRegexp)` 

**it is not necessary to replicate**
It is just the missing implementation of `IN` that causes the failure in the resolution of `OLDIPG`.    
If I comment `IN`:
image

If I don't comment `IN`:
image

that is the problem to fix.

dom-apuliasoft: I have updated with the proper implementation of the test case as suggested. For some reason I was sure that the problem was still present without the `IN` but as you correctly pointed out it is not the case. In any case, I have also extended the implementation of `executePgmCallBackTest` to look only for the errors specified (this was needed to prove we don't suffer from the data reference issue anymore). Let me know if further improvements are needed.

lanarimarco (approved):
ok

PR #545, opened 2024-06-12
Feature/ls24002974/except mock implementation

dom-apuliasoft:
## Description

Add mock implementation of `EXCEPT` statement. Differently from the other mock statements, it throw a `RuntimeException`.

Related to:
- LS24002974

## Checklist:


lanarimarco (changes requested):
The purpose of `MockStatement` and `jarikoCallback.onMockStatement` is to declare a statement that works in a mocked way; this means that the statement at runtime doesn't throw any error, but we must be aware of this.

In this case, since `EXCEPT` must throw an error at runtime, the `MockStatement` choice is an unhelpful effort. Therefore, implement `EXCEPT` as a canonical statement and throw an error in the `execute` method. Remove the new method `MockStatement.onMock` because if I want to switch some behavior based on the statement, I can do it in the default implementation of `jarikoCallback.onMockStatement`.

lanarimarco (approved):
Ok

PR #549, opened 2024-06-18
Feature/ls24002988/mock expression alloc dealloc

dom-apuliasoft:
## Description

Add support for expression mock implementations for and throw runtime error on `%alloc`, `%realloc` and `%addr` BIFs.

Related to:
- LS24002988

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/ExpressionEvaluation.kt
```diff
@@ -806,8 +810,8 @@ class ExpressionEvaluation(
         BooleanValue(enrichedDBFile.open)
     }
 
-    override fun eval(expression: SizeExpr): Value {
-        return when (expression.value) {
+    override fun eval(expression: SizeExpr): Value = proxyLogging(expression) {
+        return@proxyLogging when (expression.value) {
```
lanarimarco, line 814: I'd like to understand why in some circumstances you use a qualified return and in others, you don't.

lanarimarco (changes requested):
A small request change and I'd like a brief explanation about the usage of qualified return.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/ExpressionEvaluation.kt
dom-apuliasoft: This is needed because of a very odd behaviour I discovered rewriting `ExpressionEvaluation` with logging. As `proxyLogging` uses an inlined lambda to wrap the execution with the logging logic, when returning manually with a `return` keyword from the lambda body I found out that it returns to the caller of `proxyLogging`  instead of returning to `proxyLogging` itself, skipping execution of logging (like if we executing the `return` directly inside `proxyLogging`). Therefore the need to use a qualification after the return keyword. 

This does not happen:

- When we naturally return the value at the end of the lambda body (we don't use the `return` keyword)
- When we throw an exception

Those two represent the cases where we don't use the qualified return.
I am honestly not sure if it is a bug or intended design from kotlin as it is the first time I encounter this behaviour and found no documentation or literature of any kind mentioning it.

lanarimarco (approved):
Ok

PR #550, opened 2024-06-18
Example: display file from CLI

andraasi:
## Description

Example about using EXFMT callback to simulate IBM i display files in Jariko. The `main()` function of the example file sets up a Jariko call to a certain program (chosen with a variable, to keep it simple) with the correct configuration for display file handling. It also sets up the `onExfmt(fields: List<DSPFField>, runtimeInterpreterSnapshot: RuntimeInterpreterSnapshot): OnExfmtResponse?` callback to manage CLI interaction. Two ad-hoc example files were added: `ADD01` and `TRIM01` (with their display files).

### `ADD01`

Uses the display file format `FMT01` to ask user input for variables `A`, `B` and `PGMCTL` and write computation `A+B` to `RESULT` while `PGMCTL==0`. 

### `TRIM01`

Uses the display file format `FMT01` to ask user input for variables `STR` and `PGMCTL` and write computation `%TRIM(STR)` to `TRM` while `PGMCTL==0`. 

User input syntax is `VAR1=VALUE;VAR2=123`. A change in value is considered only for INPUT values. If `OUT` is an OUTPUT variable, prompting `OUT=123` will simply have no effect; fields whose value is to be changed are filtered to INPUT only.
Before user input is asked, previous values of INPUT and OUTPUT variables is wrote to screen. When ENTER is pressed `readln()` ends an error is displayed if input was not accepted and system requires user to re-prompt it, or will return control back to Jariko if input is accepted; calling `readln()` is blocking and will cause program to wait for user input (just like in RPG display file scenario).

## Checklist:

### DIFF
--- a/examples/src/main/kotlin/com/jariko/samples/DisplayFileFromCLI.kt
+++ b/examples/src/main/kotlin/com/jariko/samples/DisplayFileFromCLI.kt
@@ -18,8 +18,8 @@ import java.io.File
 
 // UI control functions
 
-class UnknownVariable(name: String) : Exception("Variable name $name")
-class WrongInputSyntax : Exception("Input syntax is `VAR1=VALUE;VAR2=23`")
+class UnknownVariable(name: String) : Exception("`$name`")
+class WrongInputSyntax : Exception("Should be: `VAR1=VALUE;VAR2=23`")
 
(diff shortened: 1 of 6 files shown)


lanarimarco (approved):
Ok

PR #558, opened 2024-07-03
Bugfix/ls24003187/like variable from copy

davidepalladino-apuliasoft:
## Description
This work resolves the data reference of procedure (in `D` spec) to a inline param of a subroutine declared into the same procedure, by a `/COPY`. To achieve this goal, `findSize` and `findType` of `compile_time_interpreter` have to [search both in global and local scope](https://www.ibm.com/docs/en/rdfi/9.6.0?topic=considerations-scope-definitions#hscope).

Related to #LS24003187

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
@@ -148,7 +148,7 @@ open class BaseCompileTimeInterpreter(
             if (field != null) return (field.elementSize() /*/ field.declaredArrayInLine!!*/)
         }
 
-        return findSize(rContext.statement() + rContext.subroutine().flatMap { it.statement() }, declName, conf, false)!!
+        return findSize(rContext.getStatements(), declName, conf, false)!!
     }
 
     private fun findSize(statements: List<RpgParser.StatementContext>, declName: String, conf: ToAstConfiguration, innerBlock: Boolean = true): Int? {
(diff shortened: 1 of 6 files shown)


davidepalladino-apuliasoft:
Close for rework

davidepalladino-apuliasoft:
Reopened after improvement.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
```diff
@@ -309,4 +310,18 @@ open class BaseCompileTimeInterpreter(
     private fun Parm_fixedContext.findType(conf: ToAstConfiguration): Type? {
         return this.toAst(conf, emptyList()).type
     }
+
+    private fun RpgParser.RContext.getStatements(procedureName: String?): List<StatementContext> {
```
lanarimarco, line 314: This function returns the statements in the following order:
1. The statements defined in a program
2. The statements defined in the program's subroutines
3. The statements defined in the procedure's subroutines
4. The statements defined in the procedure

I don't know if this is right or wrong, but if the goal is to give priority to the statements declared inside the procedure, maybe the order is not correct.

lanarimarco (commented):
I had some doubts about the choice to pass the procedure's name through the CompileTimeInterpreter functions, but after spending a little time understanding your work, I realized that yours has been a good choice. 

I still have a doubt regarding the implementation of RContext.getStatements, for which I need clarification.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
davidepalladino-apuliasoft: I agree with you: the order isn't correct. I found a [detailed page](https://www.ibm.com/docs/en/rdfi/9.6.0?topic=definition-scope-definitions#hsbscp) which says: "_If a local item is defined with the same name as a global data item, then any references to that name inside the subprocedure use the local definition._"

lanarimarco (approved):
Ok

PR #557, opened 2024-07-03
Bugfix/ls24003185/comptime ds extname

dom-apuliasoft:
## Description

Add support for file definitions in `CompileTimeInterpreter` to accomodate comptime resolution of DS based on EXTNAME.

Related to:
- LS24003185

## Checklist:
- [x] There are tests regarding this feature
- [x] The code follows the Kotlin conventions (run `./gradlew ktlintCheck`)
- [x] The code passes all tests (run `./gradlew check`)
- [ ] There is a specific documentation in the `docs` directory

PR #556, opened 2024-07-02
Bugfix/ls24003177/comptime definestmt

dom-apuliasoft:
## Description

Add support for `DefineStmt` in `CompileTimeInterpreter`.

Related to:
- LS24003177

## Checklist:
- [x] There are tests regarding this feature
- [x] The code follows the Kotlin conventions (run `./gradlew ktlintCheck`)
- [x] The code passes all tests (run `./gradlew check`)
- [ ] There is a specific documentation in the `docs` directory

PR #561, opened 2024-07-09
Feature/ls24003257/data struct inz hival

dom-apuliasoft:
## Description

Add support for `INZ(*HIVAL)` when applied to `DataStructValue`s.

Related to:
- LS24003257
- #556 
- #557 

## Checklist:


lanarimarco (changes requested):
In the meantime, while we are waiting for a contribution from the RPGLE developer about improving the tests, I would ask you for some small changes. In general, I prefer to expose as little as possible of Jariko, so I am safer regarding breaking changes. Therefore, the default visibility of new functions and classes should be internal.

lanarimarco (approved):
Ok

PR #575, opened 2024-07-19
Bugfix/ls24003436/copy case insensitive

davidepalladino-apuliasoft:
## Description
This work resolves an issue for resolving a name for COPY in `ETX4` file system, which is case sensitive.

Related to #LS24003436

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/facade/copy.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/facade/copy.kt
@@ -105,7 +105,7 @@ val PATTERN: Pattern = Pattern.compile("" +
 fun String.copyId(): CopyId {
     return when {
         this.contains('/') -> {
-            this.split("/,").let {
+            this.split("/", ",").let {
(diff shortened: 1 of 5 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/facade/copy.kt
```diff
@@ -106,12 +106,12 @@ fun String.copyId(): CopyId {
     return when {
         this.contains('/') -> {
             this.split("/,").let {
```
lanarimarco, line 108: This split is wrong. I have noted it right now during code review. The correct split should be: `this.split("/", ",")`.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/facade/copy.kt
```diff
@@ -106,12 +106,12 @@ fun String.copyId(): CopyId {
...
-                CopyId(it[0], it[1], it[2])
+                CopyId(it[0], it[1], it[2].uppercase(Locale.getDefault()))
```
lanarimarco, line 109: It should better capitalize all prop of CopyId not only the `member`

lanarimarco (changes requested):

lanarimarco (approved):
Ok

PR #579, opened 2024-07-26
Feature/ls24003545/caller parms

dom-apuliasoft:
## Description

Add a field to keep caller params in `InterpreterStatus`. This is immediately needed in order to fix the improper evaluation of `ParmsExpr`.

Related to:
- LS24003545

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/test/resources/VPARMSCALLEE.rpgle
+++ b/rpgJavaInterpreter-core/src/test/resources/VPARMSCALLEE.rpgle
@@ -5,44 +5,11 @@
      D §40AE           S                   LIKE(£40AE)
 
      D £40A            S             15    DIM(300)
-     D £40B            S             12    DIM(300)
-
-     D§G40DS           S                   LIKE(£G40DS)
-
-     D £G40DS          DS           500
-     D  £G40PZ                        1                                         O Lista Dinamica
-     D  £G40EX                        1                                         O Lista Estesa
-     D  £G40FM                        1                                         O Formato Ext
-     D  £G40AE                        1                                         I Utilizzo £40AE
-     D  £G40FL                        1                                         I Applica Filtro JO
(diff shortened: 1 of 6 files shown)


rpgJavaInterpreter-core/src/test/resources/VPARMSCALLEE.rpgle
```diff
@@ -0,0 +1,48 @@
+     D                SDS
+     D  £PDSPR           *PARMS
```
lanarimarco, line 2: Simplyfy this test
Since we are fixing the `*PARMS` computation issue when used in a `CALLEE` program:

- `CALLER` could pass only a param 
- `CALLEE` could have an entry with two params.

The assert in CALLEE should be  £PDSPR  = 1

For example:

**CALLER.rpgle**
```
     D P1              S             10
     
     C                   EVAL      P1='hello'
     
     C                   CALL      'CALLEE'
     C                   PARM                    P1
```

**CALLEE.rpgle**
```
     D P1              S             10
     D P2              S             10
     
     D                SDS
     D  £PDSPR           *PARMS


     C     *ENTRY        PLIST
     C                   PARM                    P1
     C                   PARM                    P2

     C     £PDSPR        DSPLY
```

lanarimarco (changes requested):
Simplify test.

lanarimarco (approved):
I made a small refactor of the tests to make them simpler and to check the number of params passed to the callee. You can take a look here: 
[5379bd1d340607a090a4aa00cc849e12aa540d24](https://github.com/smeup/jariko/pull/579/commits/5379bd1d340607a090a4aa00cc849e12aa540d24)

For the rest, it's ok.

dom-apuliasoft:
@lanarimarco well done, I missed the suggestion of asserting on the number of passed parameters in your previous comment in a first instance. Very good you noticed it on time.

PR #580, opened 2024-07-31
Display files constants fields

andraasi:
## Description

This PR wants to introduce parsing capability for display file constants fields. A constant field is specified in a display file differently from a conventional field since its declaration appears in the keywords area of the specification (columns 45-80). Classes in dspfparser module have been modified to collect keywords and extract the value of the constant field to properly place it in the previously existing object `DSPFRecord` available as attribute of the main object `DSPF` that represents a display file.  `DSPFRecord` now contains an attribute `constants` that Jariko can directly access to collect constants fields. The `onExfmt` callback now accepts a `DSPFRecord` instead of a `List<DSPFField>`. To meet the new domain object constant field `DSPFField` has now two implementers, `MutableField` and `ConstantField` each of them with its own specific properties. Both `MutableField` and `ConstantField` hold a value of type `DSPFValue` but the real object undelying is an implementer of `Value`, like `StringValue` or `DecimalValue`, for `MutableField` and `ConstantValue` for `ConstantField`.

## Checklist:


foresti-smeup:
Could you include a short summary of the changes you made in the PR description?

dspfparser/src/main/kotlin/com/smeup/dspfparser/linesclassifier/DSPFRecord.kt
```diff
@@ -9,4 +9,5 @@ import kotlinx.serialization.Serializable
 sealed interface DSPFRecord {
     val name: String
     val fields: List<DSPFField>
+    val constants: List<DSPFField>
```
lanarimarco, line 12: I noticed that both constants and fields are instances of `DSPFField`. However, `DSPFField` includes properties like `x`, `y`, and `hasError`, which seem unrelated to constants. Is this a simplification to avoid creating a separate model class, or is this the correct model to use for constants as well?

dspfparser/src/main/kotlin/com/smeup/dspfparser/linesprocessor/DSPFLine.kt
lanarimarco, line 1: I have some doubts regarding the usage of properties like:
- `isHelp`
- `isRecord`
- `isField`
- `isEtc..`

It seems that an instance of `_DSPFLine_` could potentially represent multiple categories simultaneously, such as being a help, record, and field all at once. 

For properties that are mutually exclusive, it might be better to use an enum instead of several boolean flags. This improves clarity and makes the code more maintainable, especially when the property is not private. 

This is a general suggestion, and it may also apply to other parts of the module.

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/execution/Configuration.kt
```diff
@@ -180,7 +180,7 @@ data class JarikoCallback(
     /**
      * It is invoked on EXFMT execution.
      */
-    var onExfmt: (fields: List<DSPFField>, runtimeInterpreterSnapshot: RuntimeInterpreterSnapshot) -> OnExfmtResponse? = {
+    var onExfmt: (record: DSPFRecord, runtimeInterpreterSnapshot: RuntimeInterpreterSnapshot) -> OnExfmtResponse? = {
```
lanarimarco, line 183: Explain what happens when the callback returns null.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -2394,9 +2394,9 @@ data class ExfmtStmt(
 
     override fun execute(interpreter: InterpreterCore) {
         val jarikoCallback = MainExecutionContext.getConfiguration().jarikoCallback
-        val fields = copyDataDefinitionsIntoRecordFields(interpreter, factor2)
+        val record = copyDataDefinitionsIntoRecordFields(interpreter, factor2)
         val snapshot = RuntimeInterpreterSnapshot()
-        val response = jarikoCallback.onExfmt(fields, snapshot)
+        val response = jarikoCallback.onExfmt(record, snapshot)
         response ?: error("RuntimeInterpreterSnapshot is not yet handled")
```
lanarimarco, line 2400: Change the error message as: "In the current implementation `onExfmt` callback cannot return null"

lanarimarco (changes requested):

lanarimarco (approved):

PR #590, opened 2024-08-07
Bugfix/ls24003751/elem evaluation during execution

davidepalladino-apuliasoft:
## Description
This work resolves evaluation of `%ELEM(£40A)` during execution when `£40A` is also an entry of `PLIST`.
```
     D £40A            S             15    DIM(300)
     ...
     D £40ANM          S              5S 0 INZ(%ELEM(£40A))
     ...
     C     *ENTRY        PLIST
     C                   PARM                    £40A
     C                   PARM                    £G40DS
     ...
```

In Jariko this execution caused `Void value for £40A`. More precisely, in the previous implementation `InternalInterpreter.execute` performed an initialization. `initialize` called `eval` for `NumberOfElementsExpr`, which called - at the end - `interpreterStatus.getVar`. In this case, `getVar` took the value from SymbolTable which was `NullValue`. Is wrong! Is necessary the size of array. So, the fix has been performed into `eval(expression: NumberOfElementsExpr)`, by leaving evaluation by `interpreterStatus` only, at the moment, for `OccurableDataStructureType`. 

Related to LS24003751

## Checklist:


rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00105.rpgle
lanarimarco, line 1: Add a comment in the code to highlight where the runtime error occurred. Improve the assertion, as the current one seems too weak. The "Ok" assertion only confirms that Jariko does not crash but does not ensure that Jariko behaves correctly.

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00105.rpgle
davidepalladino-apuliasoft: I decided to remove the `DSPLY` operation code because was a typo and is not necessary: without fix, with or not the `DSPLY`, there is always an runtime error. An hypothetical printing of `£40ANM` value is unnecessary because the bug is during the runtime evaluation of `£40A` parameter in `PLIST`.
So, for this reason I can't identify a strong evaluation for the Jariko behavior; I don't know how to have the evaluation result during the evaluation of that `PLIST` parameter. In this case I relied to global tests.

davidepalladino-apuliasoft:
Updated description of this PR.

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00105.rpgle
```diff
@@ -0,0 +1,35 @@
...
+    O *  `%ELEM(£40A)` for its initialization, during the evaluation
+    O *  of `£40A` `PLIST` parameter.
+     V* ==============================================================
+    O * JARIKO ANOMALY
+    O * Before the fix, the error occurred was `Void value for £40A`.
+     V* ==============================================================
+     D £40A            S             15    DIM(300)
+     D £40B            S             12    DIM(300)
+
+     D £40ANM          S              5S 0 INZ(%ELEM(£40A))                       Jariko Runtime Error: 'Void value for £40A' during evaluation of PLIST
```
lanarimarco, line 15: A comment here where you say: "runtime error during invocation of %ELEM(£40A)" is sufficient, as I think the rest does not add much to the test understanding.

lanarimarco, line 15: A comment here where you say: "runtime error during invocation of %ELEM(£40A)" is sufficient, as I think the rest does not add much to the test understanding.

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00105.rpgle
lanarimarco: Why not display the content of the variable that was not initialized? Aren't you curious to see what it is?

lanarimarco (changes requested):
The first part of the PR description is a bit too technical. Keep in mind that the first one or two sentences will be read by an RPGLE developer who might not have any knowledge of Jariko. Therefore, the language should be understandable to an RPGLE developer. 

Afterward, you can explain your solution in technical terms.

davidepalladino-apuliasoft:
Updated description of this PR.

lanarimarco (approved):

PR #591, opened 2024-08-08
Bugfix/ls24003755/ds to string when ds is greater than string

davidepalladino-apuliasoft:
## Description
This work fixes the assignment of DS content to an element of array which is String with size smaller than DS.
Under the hood has been removed restriction of size, by considering that RPG truncates the DS content if the target is smaller than source, in this case a String. Then, has been applied the fix with an important refactoring of code, by using `when` construct. If the target is `StringType`, performs the right padding or truncating based of size and varying flag. For truncating, the unused method `setSubstring` has been refactored for the purpose.
Finally, the two simple RPG programs for tests has been developed and executed on AS400.

Related to #LS24003755

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
@@ -967,7 +967,7 @@ data class DataStructValue(var value: String, private val optionalExternalLen: I
             // Check if the size of the value matches the expected size within the DS
             // TO REVIEW
             is DataStructureType -> true
-            is StringType -> expectedType.size >= this.value.length
+            is StringType -> true
             else -> false
         }
     }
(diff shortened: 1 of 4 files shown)


lanarimarco (changes requested):
The tests are a little bit complex, making it really difficult for me to understand their purpose. Please simplify them by focusing on the essential aspects or by adding comments. Always keep in mind that a good unit test should be understandable within a few seconds, without requiring knowledge of the entire context.

davidepalladino-apuliasoft:
> The tests are a little bit complex, making it really difficult for me to understand their purpose. Please simplify them by focusing on the essential aspects or by adding comments. Always keep in mind that a good unit test should be understandable within a few seconds, without requiring knowledge of the entire context.

I added an high level comment to program RPG. The program is similar to original case:
- by changing the name of variables with clear names;
- by populating each field with a proper substring of _Lorem ipsum..._.

lanarimarco (approved):

PR #592, opened 2024-08-08
Refactor/ls24003753/function call to array access recursive

dom-apuliasoft:
## Description

Given two arrays `$A` and `$B` and an index `$X`, if `$B` contains values that can be used as indices themselves, accessing an item of `$A` during an access of `$B` should be valid. For instance:
- `$A($X)` should be valid
- `$A($B($X))` should be valid as well, but it wasn't

This used to happen because the inner expression `$B($X)` was detected as a function call when contained in another array access.

To make it work `FunctionCall` to `ArrayAccessExpr` rewriting process was refactored in order to allow for recursive rewriting when the `index` itself is a `FunctionCall` that needs to be rewritten.
 
Related to:
- #589 

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00244.rpgle
+++ b/rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00244.rpgle
@@ -1,12 +1,23 @@
+      * Helper declarations
      D £DBG_Str        S             2
+
+      * Declarations needed for test purpose
      D £C6JDSO         DS           512
      D  £C6JO_TO                      2  0 DIM(10)
      D $X              S              5  0
      D §§IMPO          S              6  0
      D T$AG_I          S             21  6 DIM(15)
+
+      * Initializing array indices to avoid null value exception
      C                   EVAL      $X=2
      C                   EVAL      £C6JO_TO($X)=2
+
+      * Trying to access an array while accessing another array (like £C6JO_TO($X) inside T$AG_I).
+      * £C6JO_TO($X) should not be seen as a function call but as an array access.
+      * This should be true when this kind of array access is contained in any type expression.
      C                   EVAL      T$AG_I(£C6JO_TO($X))=
      C                             T$AG_I(£C6JO_TO($X))+§§IMPO
+
+      * Test output
      C                   EVAL      £DBG_Str='ok'
      C     £DBG_Str      DSPLY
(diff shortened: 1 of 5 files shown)


rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00244.rpgle
```diff
@@ -0,0 +1,12 @@
+     D £DBG_Str        S             2
+     D £C6JDSO         DS           512
+     D  £C6JO_TO                      2  0 DIM(10)
+     D $X              S              5  0
+     D §§IMPO          S              6  0
+     D T$AG_I          S             21  6 DIM(15)
+     C                   EVAL      $X=2
+     C                   EVAL      £C6JO_TO($X)=2
+     C                   EVAL      T$AG_I(£C6JO_TO($X))=
```
lanarimarco, line 9: What was wrong?

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00244.rpgle
dom-apuliasoft: In expressions like `T$AG_I(£C6JO_TO($X))` where the index expression is an array access it was detected as a `FunctionCall`. We already had a mechanism to replace it with an `ArrayAccessExpr` however it was only working at "first level" and was not replacing the `index` of the `index` of the `index` [...] like in line 9 of this test. Line 8 is present in order to avoid `index` being null and araising an error.

dom-apuliasoft: The PR already contains a description related to the fix. I can reword it if it is not clear enough.

lanarimarco: By "high-level description" I mean using language that allows an RPGLE coder to understand what the PR fixes or improves. This is a new rule I'd like to introduce because I realized our PR descriptions are often too technical.

The rule could be as follows:
- Highlighting the RPGLE aspects is a must.
- Implementation details, especially when the changes are significant, are highly appreciated.

For this reason I have just changed the pr template:
https://github.com/smeup/jariko/blob/develop/docs/PULL_REQUEST_TEMPLATE.md

lanarimarco (approved):
Perfect

PR #596, opened 2024-08-14
Feature/ls24003777/setll figurative constants

dom-apuliasoft:
## Description

Add support for `*START` and `*END` as factor 1 in `SETLL` statements.

Changes contain:
- The new AST creation rule
- The `StartValExpr` and `EndValExpr` definitions
- The `StartValValue` and `EndValValue`
- The `StartValType` and `EndValType` (not used yet, might be useful in future)
- The new `SETLL` rule

Related to:
- LS24003777

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/DBFileMap.kt
```diff
@@ -160,8 +160,11 @@ fun Expression.createKList(fileMetadata: FileMetadata, interpreter: InterpreterC
     return if (type() is KListType) {
         interpreter.toSearchValues(this, fileMetadata)
     } else {
-        val value = interpreter.eval(this)
-        listOf(value.asString(fileMetadata.accessFieldsType[0]))
+        when (val value = interpreter.eval(this)) {
+            is StartValValue -> listOf(fileMetadata.fields.first().fieldName)
```
lanarimarco, line 164: I don't think this is the correct implementation. If ChatGPT isn't misleading me, this code essentially means:
```
     C     *START        SETLL     TABDS01L
     C     *END          SETLL     TABDS01L
```
This moves the cursor to the beginning and then to the end of the file. However, this feature should actually be supported by Reload (an open-source project that implements record-level access APIs).

I believe the correct implementation would throw a runtime error with the message:
`*START|*END constant not yet supported.`

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/typesystem.kt
lanarimarco, line 1: For both `StartType` and `EndType`, I believe the default implementations of `hasVariableSize` and `isNumeric` are appropriate.

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/DBFileMap.kt
dom-apuliasoft: Reading the documentation again after this comment I think you are right. I apologize for the mistake, the documentation is very messy.

However we can fix easily by reverting the check on `StartValValue` and `EndValValue` and forwarding directly `*START` and `*END` to reload. At that point it should already return the runtime error you mentioned until Reload implements this feature.

dom-apuliasoft: It appears that enacting the changes I suggested in my previous comment no runtime error is thrown. This appears to be the case with every string we forward to the `setll` method. Am I missing something? Was this behaviour expected? I pushed the new code to allow for further analysis, comparison or review.

lanarimarco: You aren't missing anything.

This is the interface that Jariko uses, as defined in Reload:  
https://github.com/smeup/reload/blob/master/base/src/main/kotlin/com/smeup/dbnative/file/DBFile.kt

I believe Reload does not take care the value passed to the `setll` function, but that's not entirely wrong because `*START` and `*END` are not strings; they differ from `"*START"` and `"*END"`.

Therefore, until we decide how Reload should handle this feature, it's correct for Jariko to throw an error.

dom-apuliasoft: That is legitimate. Considering this, I am going to update the code as follow:
- Throw an error at runtime when `*START` or `*END` are with `SETLL`
- Mark the current test as ignored
- Create a clone/similar error test that expects the runtime error
- Put a note on the error test regarding why it throws an error (not implemented yet by Reload) and what to do when it gets solved

lanarimarco: I like it

lanarimarco (changes requested):
In my opinion, the mocking model should be improved. In addition to providing helper methods for loading metadata, creating, and populating the table, it should also provide the connection configuration, as it is always the same.

I reached this conclusion after analyzing the changes and noticing the following:
   - `TABDS1LDBMock`: provides an instance of `TABDS1L` metadata taken from the JSON file.
   - `createTABDS1LReloadConfig`: programmatically provides an instance of `FileMetadata` (with only two fields), even though the complete `FileMetadata` could be provided by `TABDS1LDBMock`.

In conclusion, a possible refactor could involve:
   - Moving the implementation from `MetadataDbMock` into `DBMock`.
   - Adding a default implementation of `createConnectionConfig`, similar to what you've already done in `createTABDS1LReloadConfig`.
   - Finally, implementing a `createReloadConfig` function that combines everything and can be used to create a `ReloadConfig` instance specifically for that file.

This refactor should simplify these types of tests.

dom-apuliasoft:
@lanarimarco I updated the code with the suggestion you proposed. I also added a new method `usePopulated` that is a wrapper on top of the `AutoCloseable` that executes the DB population queries on it's own as from my experience this is a step very easy to forget and difficult to trace. Let me know if you see further margin of improvement anywhere.

lanarimarco (approved):
When a colleague turns my idea into code, it makes me very happy.

PR #597, opened 2024-08-20
Bugfix/ls24003783/array in ds inz

dom-apuliasoft:
## Description

Given a DataStructure `$A` containing an array field `$B`, if `$B` was also marked with empty `INZ` keyword the system tried to initialize `$B` with a real number, throwing a runtime error.

To avoid this issue I added a check on the definition being an array during the creation process.

Related to:
- LS24003783

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/data_definitions.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/data_definitions.kt
@@ -872,26 +872,9 @@ private fun RpgParser.Parm_fixedContext.toFieldInfo(conf: ToAstConfiguration = T
     }
 
     var initializationValue: Expression? = null
-    val hasInitValue = this.keyword().find { it.keyword_inz() != null }
-    if (hasInitValue != null) {
-        if (hasInitValue.keyword_inz().simpleExpression() != null) {
-            initializationValue = hasInitValue.keyword_inz().simpleExpression()?.toAst(conf) as Expression
-        } else {
-            // TODO handle initializations for any other variables type (es. 'Z' for timestamp)
-            initializationValue = if (null != this.toTypeInfo().decimalPositions) {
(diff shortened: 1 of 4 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/data_definitions.kt
lanarimarco, line 1: Honestly, if your purpose was not to initialize the `initializedValue` in the case of fields declared as an array, I'm struggling a bit because the same commit includes both fixes and refactoring. Next time, maybe separate them into two commits, as it would be easier for me catching the nuances.

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00249.rpgle
```diff
@@ -0,0 +1,11 @@
+      * Helper declarations
+     D £DBG_Str        S             2
+
+      * Declaring an array field (£6I) in a data structure with an empty INZ keyword.
+      * Should not be assigned to a numeric
+     D £C6MIM          DS
+     D  £6I                          20P 6 DIM(99) INZ                          Importi
```
lanarimarco, line 7: What should be the value of the elements in this array when initialized like this? You should add a stronger assertion, for instance, by testing the value of the first element of the array.

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/data_definitions.kt
dom-apuliasoft: It is an oversight on my side, I was not aware I moved the code and the refactor was not intended. The only intended bit was to move `arraySizeDeclared` before the check so to be able to determine if we are considering an array or a plain value. Only now I saw the (useless) noise I generated, I apologize for that.

lanarimarco (approved):

PR #600, opened 2024-08-28
Bugfix/ls24003807/S to DS with positive number

davidepalladino-apuliasoft:
## Description
This work resolves the issue when using the `MOVE`/`MOVEL` operations to move an alphanumeric standalone field into a data structure field. The table below describes the currently handled use cases:

| Value * | DS Field Type | Precision (Scale) | Result   | Notes |
| ---------------------- | -------------- | ----------------- | ------ | ------ |
| `'00520'` | `ZONED` | 5 (0) | 520 | |
| `'00520'` | `ZONED` | 5 (2) | 5.2 | |
| `'0052 '` | `ZONED` | | Error | The value must contain only numbers. |
| `'00520'` | `PACKED` | | Error | The value must not contain only numbers. |

\* standalone field value or sub-value

### Techincal notes
Every sub-value have to match with related `DS` field. To fix these problems in Jariko, I improved:
- `DataStructValue`, that checks if a `NumberType`, as `ZONED`, has the right syntax;
- `coercing` file, that checks if the value is number and the destination RPG type is `PACKED`, to launch an Exception.
Then, `movel` simply creates an instance of `DataStructValue` by reusing the same logic for `EVAL`, allowed for `MOVE`/`MOVEL`.

Related to #LS24003807

## Checklist:
- [X] The code follows Kotlin conventions (run `./gradlew ktlintCheck`).
- [X] The code passes all tests (run `./gradlew check`).


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/coercing.kt
```diff
@@ -133,6 +134,9 @@ private fun coerceString(value: StringValue, type: Type): Value {
                             DecimalValue(BigDecimal.ZERO)
                         }
                     }
+                    type.rpgType == RpgType.PACKED.rpgType && value.value.isNumber() -> {
+                        throw UnsupportedOperationException("Cannot coerce sub-string `${value.value}` to $type.")
```
lanarimarco, line 138: We are not coercing a sub-string but a StringValue

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/coercing.kt
```diff
@@ -144,12 +148,18 @@ private fun coerceString(value: StringValue, type: Type): Value {
...
-                        DecimalValue(decimalValue)
-                    } else {
-                        val decimalValue = decodeFromDS(value.value.trim(), type.entireDigits, type.decimalDigits)
-                        DecimalValue(decimalValue)
+                    when {
+                        type.rpgType == RpgType.ZONED.rpgType -> {
+                            val decimalValue = decodeFromZoned(value.value.trim(), type.entireDigits, type.decimalDigits)
+                            DecimalValue(decimalValue)
+                        }
+                        type.rpgType == RpgType.PACKED.rpgType && value.value.isNumber() -> {
```
lanarimarco, line 156: Since we don't know how to convert a StringValue in a packed, what do you think if we throw always an error?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/coercing.kt
```diff
@@ -144,12 +148,18 @@ private fun coerceString(value: StringValue, type: Type): Value {
...
+                            throw UnsupportedOperationException("Cannot coerce sub-string `${value.value}` to $type.")
```
lanarimarco, line 157: We are not coercing a sub-string but a StringValue

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -1094,7 +1094,29 @@ data class DataStructValue(var value: String, private val optionalExternalLen: I
         } else if (data.declaredArrayInLine != null) {
             ProjectedArrayValue.forData(this, data)
         } else {
-            coerce(this.getSubstring(data.startOffset, data.endOffset), data.type)
+            val substring = this.getSubstring(data.startOffset, data.endOffset)
```
lanarimarco, line 1097: I don't undestand what issue it fixes, can you make me an example?

rpgJavaInterpreter-core/src/test/resources/ERROR39.rpgle
lanarimarco, line 1: To understand that we are testing `MOVEL` while in the `ERROR40.rpgle` you are testing `EVAL` I had to compare the files.
Add in the purposes that you are testing `MOVEL` .

rpgJavaInterpreter-core/src/test/resources/ERROR40.rpgle
lanarimarco, line 1: Add in the purposes that you are testing `EVAL`.

rpgJavaInterpreter-core/src/test/resources/ERROR41.rpgle
lanarimarco, line 1: Same issue of `ERROR39`, `ERROR40`...

rpgJavaInterpreter-core/src/test/resources/ERROR42.rpgle
lanarimarco, line 1: Same issue of `ERROR39`, `ERROR40`...

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00109.rpgle
```diff
@@ -0,0 +1,46 @@
...
+     D  A40_DS2_F3                    5  0
+     D  A40_DS2_F4                    5S 2
+
+     D A40_DS2_F4_S    S             20
+
+     C                   EVAL      A40_A50 = 'Lorem ipsum dolor si'
+     C                                         + 't amet, consectetuer'
+     C                                         + '0000500520'
+
+     C                   EVAL      A40_DS1=A40_A50                              Jariko Runtime Error: `MOVE/MOVEL not supported for the type: DataStructureType...`
```
lanarimarco, line 34: Are you sure that Jariko throws that error?

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00109.rpgle
```diff
@@ -0,0 +1,46 @@
...
+     C     A40_DS1_F1    DSPLY
+     C     A40_DS1_F2    DSPLY
+     C     A40_DS1_F3    DSPLY
+     C                   EVAL      A40_DS2_F4_S=%CHAR(A40_DS1_F4)
+     C     A40_DS2_F4_S  DSPLY
+
+     C                   EVAL      A40_DS2=A40_A50                              Jariko Runtime Error: `MOVE/MOVEL not supported for the type: DataStructureType...`
```
lanarimarco, line 41: Are you sure that Jariko throws that error?

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00109.rpgle
davidepalladino-apuliasoft: Yes, was the principal error because missed the logic in `movel`.

rpgJavaInterpreter-core/src/test/resources/smeup/MUDRNRAPU00109.rpgle
davidepalladino-apuliasoft: Yes, was the principal error because missed the logic in movel.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
davidepalladino-apuliasoft: This fixes these cases:
| Value * | DS Field Type | Precision (Scale) | Result   | Notes |
| ---------------------- | -------------- | ----------------- | ------ | ------ |
| `'00520'` | `ZONED` | 5 (0) | 520 | |
| `'00520'` | `ZONED` | 5 (2) | 5.2 | |
| `'0052 '` | `ZONED` | | Error | The value must contain only numbers. |

`data` must be `NumberType` and the syntax have to respect the rule. In this case only for "without blank characters at the end". 
You might ask me: "why this logic is not in `coerce`? Because is a rule only for DS. First of this implementation I put this logic in `coerce` and some tests failed; in other condition, out from DS, is possible to have blank at the end. In `coerce` I don't know who is the parent of the type; I know only the value and the destination type.

lanarimarco (approved):
Data structure implementation is one of the most difficult (or worst) parts of Jariko. 

This PR passes all tests, but frankly, regarding the change to the `fun get(data: FieldDefinition): Value` operator, I cannot confidently predict if this will cause regressions in real-world scenarios.

What concerns me is that we're allowing the creation of an incorrect data structure. There’s no validation in the constructor or setter methods, with validation only present in the `fun get(data: FieldDefinition): Value`. However, I don’t want to spend more time on this code review because it may not be productive, and then let's move on.

PR #601, opened 2024-09-02
Bugfix/ls24003857/bool to unlimited string coercion

dom-apuliasoft:
## Description

Add rule to convert booleans to unlimited strings. The conversion happens as it does with common strings.

> Note for RPGLE developers: Unlimited strings are a feature specific to Jariko. As the name suggests, they represent strings of unlimited size.

Here is an example of the behaviour we expect with these changes:

```
     D $A              S            512                                         This is a common string
     D $B              S               0                                        This is an unlimited string
     C                   EVAL      $A=*OFF                                      Boolean to common string conversion
     C     $A            DSPLY                                                  Should display 0
     C                   EVAL      $B=*OFF                                      Boolean to unlimited string conversion
     C     $B            DSPLY                                                  Should display 0 like $A does
```

Related to:
- LS24003857

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -454,6 +457,8 @@ data class BooleanValue private constructor(val value: Boolean) : Value {
             is BooleanValue -> value.compareTo(other.value)
             else -> super.compareTo(other)
         }
+
+    fun formatNumeric() = if (value) "1" else "0"
```
lanarimarco, line 461: This function could be private. 
My idea is to expose functions only when necessary, so I don't worry about breaking changes.

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/InterpreterTest.kt
lanarimarco, line 1: The tests related to unlimited strings are in `UNLIMIT_S.rpgle` and `UNLIMIT_DS.rpgle`. You should use those test suites.

lanarimarco (changes requested):
Add a few lines of RPGLE code as a prologue to the PR description to clarify the issue being fixed. Keep in mind that the person reading the PR description might be a lazy RPGLE developer.

lanarimarco (approved):

PR #606, opened 2024-09-09
Feature/ls24003953/parm factor1

dom-apuliasoft:
## Description

When using `PARM` with `CALL` operations, the first factor of the `PARM` can be used to specify where to put the "updated" value of the parameter when callee terminates its execution.

For instance in the following snippet, when `PROCEDURE` ends its execution, the new value of `$A` gets put in `*IN36`
```
     C                   CALL      'PROCEDURE'
     C     *IN36         PARM      *OFF          $A                1
```

This feature of `PARM` was unsupported.

From a technical standpoint changes concern:
- The addition of a `result` field in `PlistParam`
- The new AST creation rule for `PARM` factor 1
- The new assignment rule in the `execute` method of `CallStmt`
- The removal of a useless check in `misc.kt` during the creation of `PlistParam`s

Related to:
- LS24003953

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
@@ -215,7 +215,7 @@ data class StringValue(var value: String, var varying: Boolean = false) : Abstra
             else -> super.compareTo(other)
         }
 
-    fun withoutVarying() = if (varying) value.trimEnd() else value
+    internal fun trimEndIfVarying() = if (varying) value.trimEnd() else value
 
     override fun getWrappedString() = value
 }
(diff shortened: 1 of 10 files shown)


lanarimarco (changes requested):
If this feature allows to put the parameter output in more than one field, I would add in the test case another parameter.

dom-apuliasoft:
@lanarimarco test coverage should now be improved.
There are also 2 things that were added after the previous revision:
- A specific test where we check the behaviour when binding the same identifier to multiple parameters
- The `PlusExpr` evaluation rule with `StringValue` and `BooleanValue` (used in the new test cases)

All these behaviours have been checked on AS400 thanks to the support of @davidepalladino-apuliasoft

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -215,6 +215,8 @@ data class StringValue(var value: String, var varying: Boolean = false) : Abstra
             else -> super.compareTo(other)
         }
 
+    fun withoutVarying() = if (varying) value.trimEnd() else value
```
lanarimarco, line 218: I think the right name could be `trimEndIfVarying,` but that's just my point of view. 
If the name is too descriptive, we run the risk of needing to change it if the implementation changes. My approach is to declare the visibility as private or internal whenever possible.

lanarimarco (changes requested):
Another little change

lanarimarco (approved):

PR #612, opened 2024-09-17
Bugfix/ls24004074/not whole input consumed compiler directives

dom-apuliasoft:
## Description

Fix an issue where a program containing two `COPY` containing a `IF DEFINED` or `IF NOT DEFINED` directive with the same definition resulted in a wrong file output on the second one if it contained empty lines.

Related to:
- LS24004074

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/facade/compilerDirectives.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/facade/compilerDirectives.kt
@@ -33,6 +33,8 @@ val ELSE_PATTERN = Regex(""".{6}/ELSE\s*$""", RegexOption.IGNORE_CASE)
 val ENDIF_PATTERN = Regex(""".{6}/ENDIF\s*$""", RegexOption.IGNORE_CASE)
 val EOF_PATTERN = Regex(""".{6}/EOF\s*$""", RegexOption.IGNORE_CASE)
 
+private const val SIX_COLUMNS_PADDING = "      "
+
 /**
  * Resolve the EOF directive: after this directive, all rows are ignored until the end of the file
  * and marked in the result string as comments.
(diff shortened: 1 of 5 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/facade/compilerDirectives.kt
```diff
@@ -220,14 +220,11 @@ internal fun String.resolveCompilerDirectives(): String {
...
 
 private fun String.transformToComment(): String {
-    val newString: String
-    if (this.length >= 7) {
-        newString = this.substring(0, 6) + '*' + this.substring(7)
-    } else {
-        val padding = "       "
-        newString = this + padding.substring(0, 7 - this.length) + '*'
+    return when {
+        this.isBlank() -> "     *"
```
lanarimarco, line 224: "There is a missing space; the asterisk (*) must be placed in column 7. Even if your fix works, it might just be because the parser isn't perfect.

I discovered this potential issue because I didn't understand the cause of the original error. Then I modified your test to print the source passed to the parser:

```kotlin
/**
 * Duplicate definition in two different copies
 * @see #LS24004074
 */
@Test
fun executeDUPDEF() {
    val expected = listOf("ok")
    // I implemented the callback to print the source code before parsing
    val configuration = Configuration().apply {
        this.jarikoCallback.beforeParsing = { source ->
            println("source:")
            println(source)
            source
        }
    }
    assertEquals(expected, "compilerDirectives/DUPDEF".outputOf(configuration = configuration))
}
```

I hope this trick will be useful in the future."

lanarimarco (changes requested):
A very small change and a suggestion for when you address this kind of issue.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/facade/compilerDirectives.kt
dom-apuliasoft: I updated the implementation changing also the `else` branch that appeared to suffer from the opposite issue (overshooting and placing the `*` in 8th position).

PR #613, opened 2024-09-17
Bugfix/LS24004081/Z-ADD with Scalar to Array

davidepalladino-apuliasoft:
## Description
This work resolves an assignment of a **scalar value** to an **array** defined as Data Struct's field. For example:
```
     D D5COSO        E DS                  EXTNAME(D5COSO0F)
     D  D50                                DIM(99) LIKE(D$C001) INZ
     D                                             OVERLAY(D5COSO:88)

     C                   Z-ADD     1             D50
```

In addition, was analyzed and fixed an assignment of a **scalar value** to an **array** defined as Standalone. For example:
```
     D  D50            S              5  0 DIM(99) INZ

     C                   Z-ADD     1             D50
```

### Technical notes
To achieve the goal, I improved `InternalInterpreter.set` when `value` (which the source) is a scalar instead an array, under `FieldDefinition`'s branch. 

Related to #LS24004081

## Checklist:
- [x] If this feature involves RPGLE fixes or improvements, they are well-described in the summary.
- [x] There are tests for this feature.
- [x] RPGLE code used for tests is easily understandable and includes comments that clarify the purpose of this feature.
- [x] The code follows Kotlin conventions (run `./gradlew ktlintCheck`).
- [x] The code passes all tests (run `./gradlew check`).
- [ ] Relevant documentation is included in the `docs` directory.

PR #616, opened 2024-09-23
Bugfix/LS24004106/MOVEA from Scalar to Standalone array

davidepalladino-apuliasoft:
## Description
This work adds the capability to use `MOVEA` from a scalar to a Standalone array declared as:
- decimal, with decimal digits greater than number of digits of scalar;
- decimal, with decimal digits lower than number of digits of scalar;
- integer;

### Technical notes
Previous this fix, Jariko didn't take into consideration about result type when the second parameter of `MOVEA` was a scalar. For example:
```
     D A146            S             14  6 DIM(9) INZ
     C                   MOVEA     123           A146
```
It constructed the array as integer and not as decimal. More precisely, must be a decimal with number arranged from least significant digit. This problem covered `MOVE`, `MOVEL` and `EVAL` too.
So, now constructs the right array based of type.

Related to #LS24004106

## Checklist:


davidepalladino-apuliasoft:
> What I don't like about this solution is that you are addressing, what might be a more serious issue, specifically for `MOVEA`. I recall seeing the same symptoms in #613, where you fixed the assignment logic in `InternalInterpreter.set`, which I found to be a good solution.
> 
> I'm concerned that if each operation code requires its own implementation for assignment logic from scalar to vector—which should be a common behavior in Jariko—it might take years to cover all possible use cases.
> 
> For this reason, you should also try with `EVAL` and all `MOVEx` operation codes.

I agree with you. I could check with `EVAL` and `MOVEL` too.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/move_a.kt
```diff
@@ -95,6 +95,15 @@ private fun InterpreterCore.toArray(expression: Expression): ConcreteArrayValue
                 ConcreteArrayValue(mutableListOf(eval(expression)), expression.type())
             }
         }
+        is IntLiteral -> {
```
lanarimarco, line 98: What happens if type is RealLiteral

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/move_a.kt
davidepalladino-apuliasoft: While I was creating the tests, I tried this one too:
```
     D A146            S             14  6 DIM(9) INZ
     C                   MOVEA     123.5         A146
```
but didn't compile. Because the domain was for `MOVEA`, I didn't try for another operators.

lanarimarco (approved):

PR #617, opened 2024-09-24
Bugfix/ls24004056/ds string to packed

dom-apuliasoft:
## Description

This PR aims to solve many issues regarding PACKED numeric values, mainly when they are used in conjunction with DS and string values.

Here is a table summarizing new behaviours:

| Instruction type | Previous behaviour | New behaviour |
|--------|--------|--------|
| Fetch numeric type from DB file | throw error | coerce value as expected |
| Coerce a DS with packed fields to string | perform coercion with meaningless data | throw error |
| Coerce a string to a DS with packed fields | perform coercion with meaningless data | throw error | 
| MOVEL of a blank to a packed | throw error | assign `0` to the packed | 

From a technical standpoint many things have been reworked:
- Added a mechanism to override coercion in scenarios where base rules are not applicable (used to coerce DB string values to PACKED)
- Added a rule to treat blank strings as `0` when assigned to packed definitions.
- Fixed a bug where coercing a `DataStructValue` to a `StringType` resulted in returning a `DataStructValue` instead of a `StringValue`
- Fixed `LIKE` logic when applied to `DS`. It nows defines a string with length equal to the `DS` instead of cloning it.
- Add a field `fields` to `DataStructValue`. This is needed in order to keep track of the structure of the DS while performing coercions (and consequent validations).
- Added a runtime validation logic before coercion, differently from `assignableTo` method of `Value` this validation pass is meant to be run prior to the coercion, not prior to the assignment (therefore it runs before `assignableTo`).
- Added some new coercion rules
- Updated old test cases not compatible with the new changes

> NOTE: Even if there are a lot of changes, this has to be considered as a **partial release**. Many of these rules are still being tested and checked with the help of @benetti-smeup. Nevertheless it is important to release partially to unlock blocking errors at runtime.

Related to:
- LS24004056

## Checklist:

### DIFF
--- a/examples/src/main/kotlin/com/jariko/samples/RpgSamples.kt
+++ b/examples/src/main/kotlin/com/jariko/samples/RpgSamples.kt
@@ -95,7 +95,7 @@ fun passDSToJariko() {
     val dsLen = 15
     val commandLineProgram = getProgram(nameOrSource = pgm)
     // create instance
-    val dataStructValue = DataStructValue("".padStart(dsLen))
+    val dataStructValue = DataStructValue("".padStart(dsLen), emptyList())
     // set DS fields fields
     // note explicitEndOffset is required but not significant
     dataStructValue.set(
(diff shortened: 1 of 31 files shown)


dom-apuliasoft:
@lanarimarco as #618 has been merged in place of this, I think this PR can be closed without merging.

lanarimarco:
> @lanarimarco as #618 has been merged in place of this, I think this PR can be closed without merging.

Ok

PR #619, opened 2024-09-30
Bugfix/LS24004236/Indicator considered for block execution

davidepalladino-apuliasoft:
## Description
This work adds the capability to consider "indicator condition" for statements as conditional or iterator. For example:
```
     C   351             IFEQ      1
     ...
     C                   ELSE
     ...
     C                   ENDIF
```
In this case if `*IN35` is `*ON` the entire block (`IFEQ` or `ELSE`, based the condition) is executed.
For this work we have considered `IFxx`, `DOUxx`, `DOWxx`, `FOR`, `DO`, `DOU` and `DOW`.

### Technical notes
Jariko didn't cover the evaluation of indicators for statements as conditional or iterator. So, I have provided to add it by consider the logic for single statement.

Related to #LS24004236

## Checklist:


davidepalladino-apuliasoft:
Converted to draft to add test related to `DOW` and `DOU`

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/statements.kt
lanarimarco, line 1: What you've added in `statements.kt` is almost identical to what's already present in `misc.kt`.  
image
You should refactor this logic into a new function to avoid duplicated code.

lanarimarco (changes requested):
Except for the code duplication, for which I ask you refactor, I think your solution is the best it could be.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/statements.kt
davidepalladino-apuliasoft: You are right. I forgot to refactor this one and `toIndicatorCondition` too

lanarimarco (approved):

PR #620, opened 2024-09-30
LS24004159/Fix initializazion of DS without INZ

davidepalladino-apuliasoft:
## Description
On AS400 a DS numeric field not initialized is a special value. Cannot be executed any access previously initialization. For example:
```
     D DS01            DS
     D  DS01_F1                       8  2
     D  DS01_F2                      10  0
     
     ...
     C                   EVAL      MSG=%CHAR(DS01_F1)
     C     MSG           DSPLY
```
is not possible to get the value in`%CHAR`.

### Technical notes
On Jariko we have taken the decision to consider this value not as special but as `0`. Furthermore, I have fixed the wrong value as `1` when for a DS field is not specified the `INZ` keyword, both for itself and from parent.

Related to # (issue)

## Checklist:
- [ ] If this feature involves RPGLE fixes or improvements, they are well-described in the summary.


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/internal_interpreter.kt
```diff
@@ -594,8 +594,7 @@ open class InternalInterpreter(
                             it.val2,
                             value,
                             value1,
-                            value2,
-                            line
+                            value2
```
lanarimarco, line 597: Unfortunately, this change has caused a regression.  
Try, for example, to force an error in `MUTE13_01`, by replacing:

``` 
    MU* VAL1(AAA100) VAL2('910') COMP(EQ) 
```

with: 
``` 
    MU* VAL1(AAA100) VAL2('910') COMP(NE) 
```

If you run `MuteExecutionTest.executeMUTE13_01`, the error message doesn't provide useful information about the line. See below:

```
java.lang.AssertionError: Mute assertion failed - mute/MUTE13_01: Left value: "910       "  - right value: "910"
```

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/mute.kt
```diff
@@ -71,11 +71,10 @@ data class MuteComparisonAnnotationExecuted(
     val value2Expression: Expression,
     override val result: BooleanValue,
     val value1Result: Value,
-    val value2Result: Value,
-    val line: String
+    val value2Result: Value
 ) : MuteAnnotationExecuted() {
     override fun headerDescription(): String =
-        "Left value: \"${value1Result.render()}\"  - right value: \"${value2Result.render()}\" - Line $line"
```
lanarimarco, line 78: Revert change for the reasons explained earlier

lanarimarco (changes requested):

lanarimarco (approved):

PR #621, opened 2024-10-01
Bugfix/ls24004319/zero value to int

dom-apuliasoft:
## Description

Fixed a case where given a variable `$A`, if we assigned to `$A` the symbolic constant `*ZERO`/`*ZEROS` it could not be used in places where we expected a number (for instance, when passed in BIFs parameters). In these cases it is now converted to `0`.

From a technical standpoint `ZeroValue` has been extended with some conversion methods.

Related to: 
- LS24004319

## Checklist:


lanarimarco (changes requested):
### Description of PR

- What is a `*ZERO/*ZERO` symbolic constant, typo?  
- What you have fixed, considering that "coercing" is an unfamiliar term for RPGLE developers?  


### Test Readability

Your test currently covers:

- `DS` with Overlay
- `Z-ADD`
- `*ZERO`
- `EVAL`
- `%LOOKUP`
- `DSPLY`

and while this is helpful for testing multiple aspects, I think adding a simpler test that focuses solely on:

- `*ZERO`
- `EVAL`
- `DSPLY`

would make the issue clearer. For example, I would write a minimal test like this:

```rpg
     D RESULT          S              1  0
     
     C                   EVAL      RESULT = 1 + *ZERO
     C     RESULT        DSPLY     
```

dom-apuliasoft:
@lanarimarco I simplified the test case and updated the description. `*ZERO`/`*ZEROS` is not a typo, they are two different figurative constants with the same meaning like `*BLANK`/ `*BLANKS` (see [RPG figurative constants](https://www.ibm.com/docs/en/i/7.3?topic=constants-figurative)).

Based on your test I changed the `EVAL` with an addition expression to a BIF as this issue is about directly considering a `*ZERO` value in a place where we strictly expect a number. Math operations with `ZeroValue`s throw another error (because they are not supported yet) that is not blocking at the moment.

lanarimarco (approved):

PR #622, opened 2024-10-02
Feature/ls24004325/qualified like resolution

dom-apuliasoft:
## Description

Add support for declarations using `LIKE` on DS fields like following:

Given a DS `$A` with a field `$B`, declare a standalone definition `$C` using `LIKE($A.$B)`, or in RPGLE code terms

```
     D $C     S LIKE($A.$B)
```

Related to:
- LS24004325

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
@@ -237,6 +237,23 @@ open class BaseCompileTimeInterpreter(
                     }
                 }
             }
+            is QualifiedAccessExpr -> {
+                try {
+                    // Definition should be already known (DS are declared first)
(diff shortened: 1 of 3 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/compile_time_interpreter.kt
lanarimarco, line 1: Unfortunately, this module already contains too much duplicated code because, in most cases, the `size` or `elementSize` of something is strictly dependent on its `Type`. However, since we are not here to fix all of Jariko's problems, we should at least avoid adding more.

If I compare the implementation of the method you have changed, I see:
image

I think that `evaluateElementSizeOf` is nothing more than `evaluateTypeOf().elementSize()`. 

I may be wrong, but I’d like you to try this small refactor.

lanarimarco (changes requested):
Unfurtanetely,

lanarimarco (approved):

PR #624, opened 2024-10-04
chore: bump other deps

paolofacchinetti:
## Description

- Bump dependencies
- Centralize the JUnit dependency version in `grade.properties` through the new variable `junitVersion`

## Checklist:
- [] If this feature involves RPGLE fixes or improvements, they are well-described in the summary.
- [] There are tests for this feature.
- [] RPGLE code used for tests is easily understandable and includes comments that clarify the purpose of this feature.
- [] The code follows Kotlin conventions (run `./gradlew ktlintCheck`).
- [] Relevant documentation is included in the `docs` directory.

### DIFF
--- a/dspfparser/build.gradle
+++ b/dspfparser/build.gradle
@@ -32,7 +32,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-reflect:$kotlinVersion"
     testImplementation "org.jetbrains.kotlin:kotlin-test:$kotlinVersion"
     testImplementation "org.jetbrains.kotlin:kotlin-test-junit:$kotlinVersion"
-    testImplementation 'junit:junit:4.13.1'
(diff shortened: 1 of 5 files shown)


lanarimarco (changes requested):
Checklist is not mandatory, so check only what's really necessary, and in this case only the item: "The code passes all tests (run ./gradlew check).".  
Since we have dependencies included in different submodules, to avoid the issue you fixed, define the dependency versions in `gradle.properties`. For example, for JUnit, you can add an entry like this:

`junit=org.junit.jupiter:junit-jupiter-api:5.11.1`

`serializationVersion` and `kotlinVersion` are good starting points to understand how this work should be done.

I aim to improve Jariko with each iteration, and maybe this is the time to improve these aspects.

lanarimarco (approved):

PR #626, opened 2024-10-08
Bugfix/ls24004379/take projected array value

GaetanoDiNatale:
## Description

This work adds the ability to improve the sorting functionality for arrays, specifically for the case where the first factor is a reference to a field of an array in a data structure.

**Technical notes**
To achieve the goal:
- The Take function has been added to the ProjectedArrayValue class to correctly handle the values of the array.
- The sortA function has been modified to sort values in data structure's field array.

Related to #LS24004379

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/sorta.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/sorta.kt
@@ -59,9 +59,6 @@ fun sortA(value: Value, arrayType: ArrayType) {
 
             // return value
             value.container.value = resultList.joinToString("")
-
-            value.startOffset = 0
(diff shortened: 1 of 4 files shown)


lanarimarco (changes requested):
Actually, what you have done is not to fix the `SORTA` operation code in general terms, but to fix the `SORTA` operation code when the factor one is a reference to a data structure's array field. 
It is a good practice to search in the RPGLE sources already present in the repository to see if there is an example of the operation code you are working on. In this case, if you had looked at `SUBARR.rpgle`, there is a case almost identical to yours. However, the difference is substantial: in `SUBARR.rpgle` (line 40), the `A80_AR1` is a standalone array field and not a DS's array field, as in your case. Please update the PR description to take my observations into account.

lanarimarco (approved):

PR #640, opened 2024-10-18
Bugfix/LS24004504/Fix `inStatementDataDefinition` by using `API` directive

davidepalladino-apuliasoft:
## Description
This work concern `/API` directive. Is not related to vanilla RPGLE; is a feature of Jariko. For more information about this directive, read [this documentation](https://github.com/smeup/jariko/blob/develop/docs/api.md).

### Technical notes
The goal of this PR is to solve `Cannot find searched value for InStatementDataDefinition name=FOO` when `FOO` is declared as `D` spec on main program and inline from a program imported with `API` directive, like this example.

**Main program**
```
     D FOO             S             10
     C                   EXSR      SUB_R
      /API QILEGEN,PROGRAM_API
```

**Program imported**
```
     ...
     C     SUB_R         BEGSR
     C                   IF        FOO=''
     ...
     C                   ENDIF
     C                   MOVEL     'FOO'           FOO
     ...
     C                   CALL      'MYPGM'
     C                   PARM                    FOO             10
     ...
     C                   ENDSR
```
To achieve this goal, thanks of suggestion of @dom-apuliasoft to investigate `equals` of `DataDefinition`, I have implemented the missed `equals` and `hashCode` methods for `InStatementDataDefinition`. These are extremely necessary for access to `LinkedHashMap`'s entry: precisely, for `getLocal` where the searched ('data') is a `InStatementDataDefinition` but the founded is a `DataDefinition`.

Related to #LS24004504

## Checklist:


lanarimarco (approved):
Very elegant solution

PR #650, opened 2024-10-30
Bugfix/ls24004538/nested call error stack

dom-apuliasoft:
## Description

> Note: This PR does not introduce/fix any RPGLE functionality but rather fixes a regression introduced in a precedent PR.

Fix a regression introduced with changes in #644 that caused errors thrown deeply in the program stack to not cleanup the stack state similarly to what we had in #642.

Related to:
- LS24004538
- #642 
- #644 

## Checklist:


lanarimarco (commented):
I might have missed a structural problem.  
I would suggest focusing on:

- Removing `push` and `pop` in the `RpgProgram.execute` implementation
- Including them only in `CallStmt.execute` and `ProgramInterpreter.execute`

dom-apuliasoft:
I moved the stack managing logic in two places:
- When we execute a RpgProgram directly (for instance the root program)
- When we call an RpgProgram (push occurs after checking program is valid, pop occurs when the execution ends or when we quit call statement with an error indicator)

Therefore I am unmarking this for draft state for further review.

lanarimarco (approved):

PR #651, opened 2024-11-04
Bugfix/LS24004654/Use CTDATA for DS field declared as Array

davidepalladino-apuliasoft:
## Description
This work fixes the initialization of DS field, declared as array, on compile time by using content of `CTDATA`. For example:
```
     D                 DS
     DF_ARR                          42    DIM(3) CTDATA PERRCD(1)
     ...

**
*ELEMENT_1
*ELEMENT_2
*ELEMENT_3
```

### Technical notes
On Jariko the field, declared as that example, wasn't initialized with those elements. To achieve the goal, I have added:
1. logic for calculating the compile time records per line during the construction of field info, by using `CTDATA` and `PERRCD` keywords. This logic is like of `RpgParser.DspecContext.toAst`;
2. logic for initialization, in `InternalInterpreter.initialize` like the `DataDefinition` branch.

Also, the `cName` function `EndSourceContext.toAst` has been improved to evaluate the `CTDATA` in various forms. For example, with `CTDATA` and more one space between the keyword and asterisks.

Related to #LS24004654

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/data_definitions.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/data_definitions.kt
@@ -409,6 +409,8 @@ data class FieldDefinition(
         }
     }
 
+    fun isCompileTimeArray() = type is ArrayType && type.compileTimeArray()
+
     fun toDataStructureValue(value: Value) = type.toDataStructureValue(value)
 
     /**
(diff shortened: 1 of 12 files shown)


lanarimarco (changes requested):
The fix, as described in the PR description, concerns the initialization of the DS from CDATA. However, in all examples, you are testing `%LOOKUP` and assignments in a DS field of an element retrieved from the DS initialized from CDATA.  
For example, in the first test, lines [12-14] and [21-28] are, in my opinion, just noise.

davidepalladino-apuliasoft:
> The fix, as described in the PR description, concerns the initialization of the DS from CTDATA. However, in all examples, you are testing `%LOOKUP` and assignments in a DS field of an element retrieved from the DS initialized from CTDATA. For example, in the first test, lines [12-14] and [21-28] are, in my opinion, just noise.
> 
> The error in %LOOCUP I think is a side effect of wrong DS initialization from CTDATA.

Yes the side effect was exactly the wrong DS field initialization from CTDATA. I have left the real (and simplified) scenario but I could further simplify with direct access.

lanarimarco (approved):

PR #659, opened 2024-11-11
Fix error throw logic

dom-apuliasoft:
## Description

> This PR fixes internal Jariko logics. It does not add/alter any RPGLE functionality.

Fix an issue causing `tryToResolveRecursively` to throw an Exception of unexpected type.

This was specially noticeable when it was thrown while parsing an API as it caused to report errors from a `Position` that had nothing to do with the source they originated from, resulting in error reports pointing to wrong lines or out of bounds.

Related to:
- LS24004845

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/InterpreterTest.kt
+++ b/rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/InterpreterTest.kt
@@ -2549,4 +2549,19 @@ Test 6
         val expected = listOf("0", "0", "0", "1")
         assertEquals(expected, "PRSLTCALLERDUPLICATE".outputOf())
     }
+
+    @Test
+    fun missingDefinitionOnPListShouldThrowResolutionError() {
+        val systemInterface = JavaSystemInterface()
+
+        val source = """
+|     C                   CALL      'PGM'
+|     C                   PARM                    MISSING
+        """.trimMargin()
+
+        val program = getProgram(source, systemInterface)
+        assertFailsWith(AstResolutionError::class) {
+            program.singleCall(listOf())
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/InterpreterTest.kt
lanarimarco, line 1: If you change your test case as shown below:

```kotlin
@Test
fun missingDefinitionOnPListShouldThrowResolutionError() {
    // This is an inline copy with the sole purpose of adding two lines of empty code
    val copy = """
|     C*comment
|     C*comment
    """.trimMargin()

    // Re-implement the JavaSystemInterface to return the same copy content regardless of the copyId
    class MyJavaInterface : JavaSystemInterface() {
        override fun findCopy(copyId: CopyId): Copy {
            return Copy(copy)
        }
    }

    val systemInterface = MyJavaInterface()

    val source = """
|     I/COPY QILEGEN,£TABB£1DS  
|     C                   CALL      'PGM'
|     C                   PARM                    MISSING
    """.trimMargin()

    val program = getProgram(source, systemInterface)
//    assertFailsWith(AstResolutionError::class) {
        program.singleCall(listOf(), Configuration().apply { options.debuggingInformation = true })
//    }
}
```

you will see that the position reported in the stack trace is incorrect. When working with copies, the source is preprocessed, and the positions of all statements are shifted.

Generally, the `Node.error` extension function contains logic to recompute the position. However, in this case, you are passing the `CompilationUnit` instance, but there is no alternative.

So, in `resolution.kt`, you can replace your change with this:

```kotlin
val relativePosition = position?.adaptInFunctionOf(getProgramNameToCopyBlocks().second)
if (!resolved) cu.error("Data reference not resolved: ${this.name} at $relativePosition")
```

lanarimarco (changes requested):

lanarimarco (approved):

PR #664, opened 2024-11-21
Bugfix/ls24005012/stackoverflow goto

dom-apuliasoft:
## Description

Fix a stack overflow occurring when a `GOTO` was repeated a large number of times.

### Technical notes

The stack overflow occurred in specific situations where a `GOTO` was called repeatedly inside a statement which executed the `interpreter.execute` statement on its own in order to execute a subset of statements (like the `IF` statement).

This caused to spawn multiple `interpreter.execute` calls inside each other eventually resulting in a stack overflow.
This issue also caused a multiple execution of statements after the `GOTO` flow due to the fact that each of those `interpreter.execute` call resumed from the same point.

To fix these behaviours:
- The `ExecuteSubroutine` statement execution was refactored to deal with exceptions recursively.
- `GotoTopLevelException`s are now thrown directly when possible instead of being promoted from `GotoException`s. This allows to skip the promotion logic while capturing `GotoException`s simplifying management for both `GotoTopLevelException`s and  `GotoException`s.
- Strengthened tag normalization so to always be case insensitive.

Related to:
- LS24005012

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/InterpreterCore.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/InterpreterCore.kt
@@ -55,6 +55,7 @@ interface InterpreterCore {
     fun setIndicators(statement: WithRightIndicators, hi: BooleanValue, lo: BooleanValue, eq: BooleanValue)
     fun eval(expression: Expression): Value
     fun execute(statements: List<Statement>)
+    fun executeUnwrappedAt(unwrappedStatements: List<Statement>, offset: Int)
     fun dbFile(name: String, statement: Statement): EnrichedDBFile
     fun toSearchValues(searchArgExpression: Expression, fileMetadata: FileMetadata): List<String>
     fun fillDataFrom(dbFile: EnrichedDBFile, record: Record)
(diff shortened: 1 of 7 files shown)


lanarimarco (changes requested):
Why didn’t you implement all aspects related to `GOTO` handling in `ExecuteSubroutine.execute`?

dom-apuliasoft:
I didn't implement all the `GOTO` related logic in `ExecuteSubroutine.execute` because in that case when we catch a `GotoException`, we would need to call `interpreter.execute` again doing something like:
```
interpreter.execute -> ExecuteSubroutine -> interpreter.execute -> ExecuteSubroutine -> ...
```

This is what was causing the stack overflow in first place.

In that scenario, throwing a a `GotoException` means going back in the call stack up to the `ExecuteSubroutine` that threw it and spawning new executes from there.

Moving this logic to the interpreter means going back to the original `interpreter.execute` no matter what and always spawning a single new execute *while stopping the current one*. This makes sure that when the `GOTO` subroutine finish, we don't fallback to a parent subroutine execute logic we started somewhere in the `ExecuteSubroutine` call stack and of which we lost track and control.

dom-apuliasoft:
@lanarimarco actually you where right. I moved the logic in `ExecuteSubroutine` and with minor adjustments it seems to cause no harm. My previous comment must be based on an oversight on my side so please ignore it.

lanarimarco (approved):

PR #665, opened 2024-11-22
bugfixLS24005040/Interrupt Jariko and improvements for `Z-ADD`

davidepalladino-apuliasoft:
## Description
This work is started by @lanarimarco as improvement for interruption of execution if a current thread is dead. While he was working to this with a little refactoring for `IntValue`, several `Z-ADD` tests fell apart. The reason was related to the missed truncation for a number (source) with number of digits greater than destination used as target in `Z-ADD`. For example:
```
     D SRC             S              6  0 INZ(241122)
     C                   Z-ADD     SRC           RES               4 0
```

## Technical notes
On Jariko this caused `241122 cannot be assigned to RES of type NumberType`. To apply the fix, I have improved the assignment by replacing the direct utilization of `assign` in `ZAddStmt.execute` with a function located in the new `zadd.kt`. This function provides the complex operations for `Z-ADD` as the truncation of number in these scenarios:
- integer -> integer;
- decimal -> decimal;
- integer -> decimal;
- decimal -> integer.

This is possible only when `ZAddLegacyFlag` is set to `On`. Otherwise, Jariko raises an exception where a source value `cannot be assigned` to a target with its size is lower.

Related to #LS24005040

## Checklist:


lanarimarco (changes requested):
Good work, but since I don’t like an operator that allows assigning values greater than the target variables, I think it’s better to manage this behavior as a feature flag, `ZAddLegacy` (default: false).  
If this feature flag is false, the Z-ADD behaves as before.

To implement a feature flag:

- Define the `FeatureFlag.ZAddLegacy` enum.  
- Add the `IFeaturesFactory.isZAddLegacy` function.  
- Invoke `IFeaturesFactory.isZAddLegacy` wherever necessary.  
- Add documentation in this section: [Enabling Experimental Features](https://github.com/smeup/jariko/blob/develop/docs/development.md#enabling-experimental-features)

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/smeup/MULANGT02ConstAndDSpecTest.kt
```diff
@@ -291,6 +294,8 @@ open class MULANGT02ConstAndDSpecTest : MULANGTTest() {
      */
     @Test
     fun executeMUDRNRAPU00206() {
+        smeupConfig = turnOnZAddLegacyFlagConfig
```
lanarimarco, line 297: In my opinion, this assignment is unhelpful. You should pass the instance of `Configuration` returned by `turnOnZAddLegacyFlagConfig` to the `outputOf` function.
Follow this suggestion also for the other methods where you used this pattern.

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/smeup/MULANGT02ConstAndDSpecTest.kt
```diff
@@ -19,6 +19,9 @@ open class MULANGT02ConstAndDSpecTest : MULANGTTest() {
 
     @AfterTest()
     override fun tearDown() {
+        // Turning off ZAddLegacy in case some test has activated it.
+        smeupConfig = turnOnZAddLegacyFlagConfig
```
lanarimarco, line 23: This is not necessary if you proceed as explained later.

lanarimarco (approved):

PR #673, opened 2024-12-03
Bugfix/ls24005170/goto branching quit

dom-apuliasoft:
## Description

Fix an issue that caused `GOTO` pointing inside statements with branches not to quit the statement after the corresponding branch execution was finished.

### Technical notes

The issue was caused by the fact that after the statement unwrapping (needed to perform `GOTO` jumps in nested operations without particular overhead, see #643) we lost cognition of where each branch teminated, treating each subsequent instruction as a standalone non-scoped instruction.

To address this issue:
- The concept of unwrapping was separated from just doing an explode on statements
- A dedicated `UnwrappedStatementData` structure was created to enforce clean and understandable interfaces when dealing with unwrapped statements
- A new utility `List<Statement>.unwrap()` method was added (with a behaviour similar to `List<Statement>.explode()` but with custom logics)
- A new `CustomStatementUnwrap` interface was added to allow custom unwrap behaviour for inheriting statements. There should really be only 2 (`IF`, `SELECT`) but this mechanism allow for easier debug, more coincise code, separation of concerns and easier maintenance in case with find further issues. All for a very low cost.

Related to:
- LS24005170
- #643

## Checklist:
- [x] Relevant documentation is included in the `docs` directory.

### DIFF
--- a/docs/goto.md
+++ b/docs/goto.md
@@ -128,3 +128,8 @@ already execute their child on their own. Increasing the index by one instead of
 
 Eventually each `CompositeStatement` calls the base `execute(statements: List<Statement>)` method, therefore this method has to be
 imagined like a middleware to keep the context of `GOTO` calls sane and consistent called only when a `GOTO` exception is caught.
+
+## Next steps
+
+This implementation is very complex and has many nuances that are very hard to understand. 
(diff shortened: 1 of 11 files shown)


lanarimarco (approved):

lanarimarco:
Add doc

PR #595, opened 2024-08-13
Bugfix/ls24003380/immutable base not expression

dom-apuliasoft:
## Description

In expressions like `NOT($A($X))` where `$A` is an array and `$X` is an index the program crashed because the expression `$A($X)` was firstly detected a function call and then converted in an array access.

This conversion failed because of a mutability issue.

For this reason the `base` field of `NotExpr` is now mutable.

Related to:
- #559 
- #589 
- #574 

## Checklist:
- [x] If this feature involves RPGLE fixes or improvements, they are well-described in the summary.
- [x] There are tests regarding this feature
- [x] RPGLE code used for tests is easily understandable and includes comments that clarify the purpose of this feature.
- [x] The code follows the Kotlin conventions (run `./gradlew ktlintCheck`)
- [x] The code passes all tests (run `./gradlew check`)
- [ ] There is a specific documentation in the `docs` directory

PR #676, opened 2024-12-04
Bugfix/ls24005243/checkr factor2 expression

dom-apuliasoft:
## Description

Allow support for expressions in factor 2 of  `CHECK` and `CHECKR` after the `:`. Execution logic remains the same.

Related to:
- LS24005243

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -854,9 +854,9 @@ data class CheckStmt(
 
 @Serializable
 data class CheckrStmt(
-    val comparatorString: Expression, // Factor1
-    val baseString: Expression,
-    val start: Int = 1,
+    var comparatorString: Expression, // Factor1
```
lanarimarco, line 857: I think this property, along with the ones below, should remain immutable as they were before.

lanarimarco (changes requested):
I think, it could be helpful fixes this issue also for CHECK statement.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
dom-apuliasoft: This is needed to prevent mutability errors when expression is a function call or an array access expression. Here is an example of the very same issue fixed previously. there are many more identical to this in PR history: https://github.com/smeup/jariko/pull/595

dom-apuliasoft: @lanarimarco while fixing for `CHECK` statement, I noticed that there is a `FactorContext.toIndexedExpression()` method that seems to do exactly what the new method `FactorContext.toPositionalExpression()` does with only support for literals. Currently it is only used for `SCAN` and `XLATE` statements. Do you think it is safe to state that `FactorContext.toIndexedExpression()` tries to do what `FactorContext.toPositionalExpression()` is now doing?

In that case I think it would be wise to drop the old method and replace it with the new one while keeping the old name. This should provide more clarity and avoid having duplicate symbols with different logic whose use is ambiguous.

lanarimarco: > This is needed to prevent mutability errors when expression is a function call or an array access expression. Here is an example of the very same issue fixed previously. there are many more identical to this in PR history: #595

Ah, okay, I remember now.  
Can you add a test? This will make the reason for your change clearer.  

Thank you.

lanarimarco: > In that case I think it would be wise to drop the old method and replace it with the new one while keeping the old name. This should provide more clarity and avoid having duplicate symbols with different logic whose use is ambiguous.

Absolutely agree.

lanarimarco (changes requested):
Okay, I agree with your proposals.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -817,9 +817,9 @@ data class SetgtStmt(
 
 @Serializable
 data class CheckStmt(
```
lanarimarco, line 819: Update pr description

lanarimarco (approved):

PR #677, opened 2024-12-05
Bugfix/ls24005268/ds dim on constant

dom-apuliasoft:
## Description

Add support for DS fields defined with `DIM` keyword based on a constant.

### Technical notes

From a technical stand point this fix should generally improve the definition of constants.
Constants are now processed in their own step before everything else in isolation from the rest instead of relying on the statement sorting.

This way we are sure:
- We don't try to define constants multiple times
- Constants are always available when referenced by other definitions

Related to:
- LS24005268

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/utils/misc.kt
```diff
@@ -178,4 +178,8 @@ internal fun Node.getContainingCompilationUnit() = ancestor(CompilationUnit::cla
 
 internal fun List<UnwrappedStatementData>.indexOfTag(tag: String) = indexOfFirst {
     it.statement is TagStmt && (it.statement as TagStmt).tag.lowercase() == tag.lowercase()
+}
+
+internal inline fun <T, R> Iterable<T>.mapNotNullOrError(transform: (T) -> R?): List<R> {
+    return this.mapNotNull { kotlin.runCatching { transform(it) }.getOrNull() }
```
lanarimarco, line 184: I'm aware that we have inherited this kind of error handling, and possibly it is intentional, but just out of curiosity, what kinds of errors are caught during that phase?

lanarimarco (commented):
Ok I have only a question about run catching-

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/utils/misc.kt
dom-apuliasoft: I am not 100% sure either but as far as I am aware the main reasons I found while dealing with it are the following:
- When AST phase fails, syntax error are reported in the erroring site. In this case `runContext` allows to filter failing definitions without blocking while recording errors at the same time.
- As in the past we had multiple subsequent definition steps/passes it also allowed to retry definition in a later step. For instance if we could not resolve a DS referencing some other definition in the first pass, it could be retried in the second pass. However, this should not be the case anymore.

There might be more reasons I am not aware of. This are the ones I encountered on my own.

lanarimarco (approved):

PR #683, opened 2024-12-10
Bugfix/LS24005329/Improve comparison between `*HIVAL` and String

davidepalladino-apuliasoft:
## Description
This work is an improvement for PR #680, by using other comparison operators between `*HIVAL` and String.

Related to #LS24005329

## Checklist:


lanarimarco (approved):
The implementation is very clean and closely aligns with what I envisioned to address these issues. However, the tests could be clearer:  
- You are using a D-spec of type array, which could lead to the impression that we are fixing something related to arrays.  
- The names of the D-specs are `ARR01` and `ARR02`.  

In my opinion, the use of arrays is unnecessary, and the variable names should be more meaningful. For instance, instead of `ARR01` or `ARR02`, consider using `NOHIVAL` and `HIVAL`.  

Please take these suggestions into account for future implementations.

PR #693, opened 2025-01-16
Bugfix/LS25000142/Fix DS access between `QUALIFIED` and not

davidepalladino-apuliasoft:
## Description
This work solves the DS field access based of `QUALIFIED` keyword declared as explicit or not (by using `LIKEDS`). For first case already exists a test, `parseMUTE12_01_ast`. For second case there is a test like this snippet:
```
     D ML            E DS                  EXTNAME(MULANGTL) INZ
     D MLL             DS                  LIKEDS(ML)

     C                   EVAL      MLSYST='IBMI'
     C     MLSYST        DSPLY
     C     MLL.MLSYST    DSPLY
     ...
```
For more explanation, from this code we have:
- `ML` which is not `QUALIFIED`;
- `MLL` which is `QUALIFIED` by using `LIKEDS` even if the parent DS (`ML`) is not `QUALIFIED` (see [this](https://www.ibm.com/docs/en/i/7.4?topic=keywords-likedsdata-structure-name) for more information) .

### Technical notes
On Jariko there was one problem by considering that snippet: logic of _dot notation_ was forced ignored. During `MLL.MLSYST` evaluation was considered `ML` instead `MLL`. With this fix has been implemented:
- optional _dot notation_ for not `QUALIFIED` Data Structure;
- mandatory _dot notation_ for `QUALIFIED` Data Structure.

However, the resolution of a field used like in this snippet:
```
     C     MLSYST        DSPLY
```
is done only for not `QUALIFIED` Data Structure.

Related to #LS25000142

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/ast/statements.kt
```diff
@@ -960,7 +960,14 @@ data class CallStmt(
                     )
                 }
             }
-            targetProgramParams[index].name to interpreter[it.result.name]
+            /*
+             * TODO: On RPGLE is possible to have nested field with an access like this: ds.field.subfield.
+             *  Consider to refactor this logic when is necessary to add subfield capability.
+             */
+            val resultName = if (it.result.name.contains("."))
```
lanarimarco, line 967: Remove the TODO and replace it with: If there are two dots, we should throw an error. This makes it clear that this feature is not supported and helps avoid subtle errors.

lanarimarco (changes requested):
Very good work.  
To complete it, in addition to the inline proposals:  
- Add a test with a DS declared as `QUALIFIED`.  
- Add an error test to verify that Jariko allows access to fields without dot notation only for `UNQUALIFIED` DS.

davidepalladino-apuliasoft:
> Very good work. To complete it, in addition to the inline proposals:
> 
>     * Add a test with a DS declared as `QUALIFIED`.
> 
>     * Add an error test to verify that Jariko allows access to fields without dot notation only for `UNQUALIFIED` DS.

Thanks @lanarimarco, I appreciate it!
I can provide an error test for proving access with dot notation only for `UNQUALIFIED` DS but I found several tests which utilize that keyword:
- `PROCEDURE_N` with `MYDS`;
- `MUTE12_01` (of `data.ds` folder) with `DS1`;
- `MUTE12_01` (of `overlay` folder) with `DS1`;
- `MU108006` with `FileDS`;
- `T10_A60_P04-P07` with `A60PDS`.

All of these have the _dot notation_ utilization. Although, do you want another test under `smeup` folder as `MUDRNRAPU00191` (the next one)?

lanarimarco:
> All of these have the _dot notation_ utilization. Although, do you want another test under `smeup` folder as `MUDRNRAPU00191` (the next one)?

You are right,  ignore my request.

lanarimarco (approved):

PR #705, opened 2025-02-06
Bugfix/LS24003795/Incongruous definifion between two DS by using API

davidepalladino-apuliasoft:
## Description
This work is related to Jariko, for the logic of `/API` directive.

### Technical notes
We consider this program:
```
      /COPY QILEGEN,£PDS
      /COPY QILEGEN,£C5PDS

      /API QILEGEN,MUDRNRAPU001101_API      
     ...
```
and the program added with `/API` directive:
```
     D/COPY QILEGEN,£PDS
```
Under the hood we have, from main program, a DS (by `£PDS`) with another fields (by `£C5PDS`); on the other hand, the program imported by using `API` declares the same DS (by `£PDS`) without additions. So, the fix has been made thanks to these assertions:

- `API` program still works with its definition because the same DS (without additions) is declared from main program (with additions);
- the DS enhanced of main program still works because is the only one considered during the execution.

This is a work of 22nd August 2024.

Related to #LS24003795

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
```diff
@@ -2225,18 +2225,21 @@ internal fun <T : AbstractDataDefinition> List<T>.removeDuplicatedDataDefinition
 }
 
 internal fun AbstractDataDefinition.matchType(dataDefinition: AbstractDataDefinition): Boolean {
-    fun Type.matchType(other: Any?): Boolean {
-        // TODO: Improve logic for StringType/UnlimitedStringType matching
+    fun Type.matchType(called: Any?): Boolean {
```
lanarimarco, line 2228: Why did you replace `other` with `called`?
In my opinion `called` is not more expressive, and maybe also not really correct, because you are matching the type of receiver with an`other` type.

davidepalladino-apuliasoft: Yes, I'll change it.

lanarimarco (approved):

PR #702, opened 2025-02-04
Develop before revert/LS25000520/Fix concurrence between DS and File

davidepalladino-apuliasoft:
## Description
This work is related to #696 . In addition:
- `removeFileFieldsDefinedInDs` has been removed because there is a new logic made from #700 with `removeStandaloneDefinedAsFieldOfUnqualifiedDs`;
- a new test has been added where the file in `EXTNAME` is different from `F` spec but shares same fields.

Related to #LS25000520

## Checklist:
- [x] If this feature involves RPGLE fixes or improvements, they are well-described in the summary.
- [x] There are tests for this feature.
- [x] RPGLE code used for tests is easily understandable and includes comments that clarify the purpose of this feature.
- [x] The code follows Kotlin conventions (run `./gradlew ktlintCheck`).
- [x] The code passes all tests (run `./gradlew check`).
- [ ] Relevant documentation is included in the `docs` directory.

PR #707, opened 2025-02-14
Bugfix/LS25000845/Fix twice definition between field defined both in `F`  and like `S`/`DS`

davidepalladino-apuliasoft:
## Description
This work resolves the definition, as `S` or `DS`, of a variable already defined as File field, like these snippets:
```
     FB£G11G0V  IF   E           K DISK
     D S$INTE          DS            22
     D  S4INTE                        4    OVERLAY(S$INTE:1)
```
and
```
     FB£G11G0V  IF   E           K DISK
     D S$INTE          S             22
```
In this case, `B£G11G0V` defines `S$INTE` but there is a redefinition of same variable as Standalone or Data Structure.

## Technical notes
On Jariko this caused `S$INTE has been defined twice`. So I have:
- added a discriminant for `DataDefinition`, if is created from file;
- improved `addIfNotPresent` by checking if the _actual_ is `fromFile` and `StringType`, and _source_ is `DataStructType` or `StringType` with same size. In this case is legal to _rewrite_ the data definition.  

For the last point, why to rewrite? There are two reasons by considering the first snippet (where we have a redefinition as DS):
- the work of #702;
- the certain ability from program to access to `S4INTE` field instead `S$INTE` parent.

Related to #LS25000845

## Checklist:


lanarimarco (changes requested):

lanarimarco (changes requested):
Here’s the corrected version while keeping the format unchanged:  

I am not sure we have caught the real origin of the issue.  
In my opinion, the problem is: if `A` is a `DS` and `B` is a `D` spec alphanumeric with the same length as `A` (`A` and `B` have the same name),  we need to consider `A` and remove `B` because `B` can be replaced with `A` but not vice versa, regardless of whether `A` and `B` are coming from a file.  

In your work, I see a first attempt where you evaluate the new property `DataDefinition.fromFile`, followed by an adjustment concerning a refactor around the data definition providers, which depends on the order in which the collection of data definition providers is created.  

My doubt is: if I have an RPG program and I declare `A` and `B` as explained above, I think this fix does not work.

Try the proposed use case in as400 and let me know.

davidepalladino-apuliasoft:
> I am not sure we have caught the real origin of the issue. In my opinion, the problem is: if `A` is a `DS` and `B` is a `D` spec alphanumeric with the same length as `A` (`A` and `B` have the same name), we need to consider `A` and remove `B` because `B` can be replaced with `A` but not vice versa, regardless of whether `A` and `B` are coming from a file.
> 
> In your work, I see a first attempt where you evaluate the new property `DataDefinition.fromFile`, followed by an adjustment concerning a refactor around the data definition providers, which depends on the order in which the collection of data definition providers is created.
> 
> My doubt is: if I have an RPG program and I declare `A` and `B` as explained above, I think this fix does not work.
> 
> Try the proposed use case in as400 and let me know.

I tried on AS400 with this code:
```
     D DEF             DS            22     
     D DEF             S             22     

     
     C                   SETON                                        RT
```
and doesn't compile it with this (in Italian) error: 
```
 *RNF3316 30      1 La voce è stata già definita su una specifica di definizione;
                    specifica ignorata.
```

Also,

lanarimarco:
> > I am not sure we have caught the real origin of the issue. In my opinion, the problem is: if `A` is a `DS` and `B` is a `D` spec alphanumeric with the same length as `A` (`A` and `B` have the same name), we need to consider `A` and remove `B` because `B` can be replaced with `A` but not vice versa, regardless of whether `A` and `B` are coming from a file.
> > In your work, I see a first attempt where you evaluate the new property `DataDefinition.fromFile`, followed by an adjustment concerning a refactor around the data definition providers, which depends on the order in which the collection of data definition providers is created.
> > My doubt is: if I have an RPG program and I declare `A` and `B` as explained above, I think this fix does not work.
> > Try the proposed use case in as400 and let me know.
> 
> I tried on AS400 with this code:
> 
> ```
>      D DEF             DS            22     
>      D DEF             S             22     
> 
>      
>      C                   SETON                                        RT
> ```
> 
> and doesn't compile it with this (in Italian) error:
> 
> ```
>  *RNF3316 30      1 La voce è stata già definita su una specifica di definizione;
>                     specifica ignorata.
> ```
> 
> So, the problem to resolve is the capability to replace a Field of File with the new declared from program if the size and name are equals.

RPGLE is surprisingly strange, ok thanks for your test.

lanarimarco (dismissed):

davidepalladino-apuliasoft:
> > > I am not sure we have caught the real origin of the issue. In my opinion, the problem is: if `A` is a `DS` and `B` is a `D` spec alphanumeric with the same length as `A` (`A` and `B` have the same name), we need to consider `A` and remove `B` because `B` can be replaced with `A` but not vice versa, regardless of whether `A` and `B` are coming from a file.
> > > In your work, I see a first attempt where you evaluate the new property `DataDefinition.fromFile`, followed by an adjustment concerning a refactor around the data definition providers, which depends on the order in which the collection of data definition providers is created.
> > > My doubt is: if I have an RPG program and I declare `A` and `B` as explained above, I think this fix does not work.
> > > Try the proposed use case in as400 and let me know.
> > 
> > 
> > I tried on AS400 with this code:
> > ```
> >      D DEF             DS            22     
> >      D DEF             S             22     
> > 
> >      
> >      C                   SETON                                        RT
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and doesn't compile it with this (in Italian) error:
> > ```
> >  *RNF3316 30      1 La voce è stata già definita su una specifica di definizione;
> >                     specifica ignorata.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So, the problem to resolve is the capability to replace a Field of File with the new declared from program if the size and name are equals.
> 
> RPGLE is surprisingly strange, ok thanks for your test.

You're welcome!

lanarimarco (commented):
Resolve conflicts.
Out of curiosity, since you are the only one currently working on Jariko, I don't understand why this conflict occurred.

davidepalladino-apuliasoft:
> Resolve conflicts. Out of curiosity, since you are the only one currently working on Jariko, I don't understand why this conflict occurred.

Done!
Because a test has been moved from a suite to another.

lanarimarco (approved):

PR #712, opened 2025-03-10
Bugfix/ls25001162/zadd binary

dom-apuliasoft:
## Description

Adjust logic applied to binary fields that caused an error on reassignments

## Technical notes

The issue was caused by an improper assignment of the size when binary numbers where applied to fields.

In this case binary fields should behave like follows:
- Precision 1 to 4: 2 bytes in size
- Precision 5 to 9: 4 bytes in size

To simulate this behaviour we used the digits to determine its size, however:
- This was not right because assigning correct values could arise not assignable errors (for instance precision 4 assigning value 8224).
- The size was already properly retrieved accordingly when calling the `size` getter making the previous logic an unnecessary duplicate.

Related to:
- LS25001162

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/test/resources/overlay/MUTE12_02.rpgle
+++ b/rpgJavaInterpreter-core/src/test/resources/overlay/MUTE12_02.rpgle
@@ -277,15 +277,15 @@
      C                   EVAL      DS0010=*HIVAL
     MU* VAL1(DS0010) VAL2(-9999999999,99999) COMP(EQ)
      C                   EVAL      DS0010=*LOVAL
-      * TODO incorrect, DS0011 is 2 bytes
+      * TODO incorrect, DS0011 is 2 bytes with precision 2
     MU* VAL1(DS0011) VAL2(99) COMP(EQ)
      C                   EVAL      DS0011=*HIVAL
     MU* VAL1(DS0011) VAL2(-99) COMP(EQ)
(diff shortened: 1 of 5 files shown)


rpgJavaInterpreter-core/src/test/resources/overlay/MUTE12_02.rpgle
```diff
@@ -277,15 +277,15 @@
      C                   EVAL      DS0010=*HIVAL
     MU* VAL1(DS0010) VAL2(-9999999999,99999) COMP(EQ)
      C                   EVAL      DS0010=*LOVAL
-      * TODO incorrect, DS0011 is 2 bytes
+      * TODO incorrect, DS0011 is 2 bytes with precision 2
```
lanarimarco, line 280: I don't understand this comment, and in general, all comments classified as `TODO incorrect`.

dom-apuliasoft: To be fair, I don't fully understand them either and I decided to keep them only because I was not sure. I imagine it meant the check was only looking for the value and not the size but this idea is weak as this check is done anyway kotlin side. If you agree with me, I think we can safely remove them.

lanarimarco: Try to involve Benetti

lanarimarco (approved):

PR #718, opened 2025-04-03
Bugfix/LS25001579/Array of DS as param of `CALL`

davidepalladino-apuliasoft:
## Description
This work resolves the utilization of DS field, declared as array, as parameter of `CALL`, like in these snippets:
##### Main program
```
     D DS1             DS
     D  DS1_F1                        5  0 DIM(3) INZ

     ...
     C                   CALL      `OTHER_PGM`
     C                   PARM                    DS1_F1
```

##### Called program
```
     D VAR_PARM        S             15

     C     VAR_PARM      DSPLY

     C     *INZSR        BEGSR
     C     *ENTRY        PLIST
     C                   PARM                    VAR_PARM
     C                   ENDSR
```
### Technical notes
On Jariko missed the implementation for `asString` and `takeAll`, on `ProjectedArrayValue`. `takeAll` constructs, as string, the concatenation of numbers by considering the size and scale of declared array, as well as on AS400.

Related to #LS25001579

## Checklist:
- [x] If this feature involves RPGLE fixes or improvements, they are well-described in the summary.
- [x] There are tests for this feature.
- [x] RPGLE code used for tests is easily understandable and includes comments that clarify the purpose of this feature.
- [x] The code follows Kotlin conventions (run `./gradlew ktlintCheck`).
- [x] The code passes all tests (run `./gradlew check`).
- [ ] Relevant documentation is included in the `docs` directory.

PR #719, opened 2025-04-08
Bugfix/ls25001659/nope to array type

dom-apuliasoft:
## Description

Fix a corner case causing the interpreter to throw an unexpected error when exiting from a called program returning a DS in a compatible array.

### Technical notes

This issue arised after #718 and its investigation lead to undercover some structural flaws, mainly in `coerce`.
The case is similar to `MUDRNRAPU001107_P` and `MUDRNRAPU001108_P` but only occurs when the value being returned is a `DataStructValue` and not a `StringValue`.

Changes are the following:
- Add an overload for `render` in `DataStructValue` to improve error output.
- Tweak `StringType.assignableTo` logic. Strings can be assigned to an arrayValue of same size and to any other string.
- Fix `DataStructValue` copy logic. Previously it did not copy the `unlimitedStringField`.
- Add specialized `coerce` logic for `DataStructValue`.

Related to:
- LS25001659

Depends on:
- https://github.com/smeup/jariko/pull/718

## Checklist:

### DIFF
--- a/kolasu/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
+++ b/kolasu/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
@@ -73,8 +73,18 @@ data class ReferenceByName<N>(val name: String, var referred: N? = null) where N
         get() = referred != null
 }
 
+@Deprecated(
+    message = "Deprecated for performance reasons",
+    replaceWith = ReplaceWith("tryToResolve(candidates: Map<String, N>, caseInsensitive: Boolean = false)")
+)
 fun <N> ReferenceByName<N>.tryToResolve(candidates: List<N>, caseInsensitive: Boolean = false): Boolean where N : Named {
     val res = candidates.find { if (it.name == null) false else it.name.equals(this.name, caseInsensitive) }
+    this.referred = res
+    return res != null
+}
+
+fun <N> ReferenceByName<N>.tryToResolve(candidates: Map<String, N>, caseInsensitive: Boolean = false): Boolean where N : Named {
+    val res = if (caseInsensitive) candidates[this.name.uppercase()] else candidates[this.name]
     this.referred = res
     return res != null
 }
--- a/kolasu/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/kolasu/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -1,6 +1,23 @@
+/*
+ * Copyright 2019 Sme.UP S.p.A.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package com.strumenta.kolasu.model
 
 import java.util.LinkedList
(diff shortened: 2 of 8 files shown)


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
```diff
@@ -86,7 +86,7 @@ data class StringValue(var value: String, var varying: Boolean = false) : Abstra
 
     override fun assignableTo(expectedType: Type): Boolean {
         return when (expectedType) {
-            is StringType -> expectedType.length >= value.length.toLong()
+            is StringType -> true
```
lanarimarco, line 89: If you run this program:

```
     D AAAAA           S              1
     D BBBBB           S              2
     
     C                   EVAL      BBBBB='BB'
     C                   EVAL      AAAAA=BBBBB
     C     AAAAA         DSPLY     
```

Already is displayed B, and this means that `assignableTo` in some cases is not invoked.
Which problem does this change fix?

rpgJavaInterpreter-core/src/test/kotlin/com/smeup/rpgparser/evaluation/InterpreterTest.kt
```diff
@@ -2605,9 +2605,9 @@ Test 6
 
     @Test
     fun executeBIG_DO_LOOP() {
```
lanarimarco, line 2607: Possibily merging error, copy the actual version from the develop.
This version was wrong because the failure in a thread execution is only logged but the test is not marked as failed. 

The correct version is this:

```kotlin
    @Test
    fun executeBIG_DO_LOOP() {
        var error: Throwable? = null
        val jarikoExecutorThread = Thread {
            error = assertFails { executePgm(programName = "BIG_DO_LOOP") }
        }
        val jarikoKillerThread = Thread {
            println("Waiting 2 seconds before killing jariko")
            Thread.sleep(2000)
            println("Killing jariko")
            jarikoExecutorThread.interrupt()
            println(jarikoExecutorThread.state)
        }
        jarikoExecutorThread.start()
        jarikoKillerThread.start()
        jarikoExecutorThread.join()
        require(error != null) { "Jariko should throw an error" }
        require(getRootCause(error!!) is InterruptedException)
    }
```

rpgJavaInterpreter-core/src/test/resources/ARRAYDS.rpgle
```diff
@@ -0,0 +1,15 @@
...
+     V* ==============================================================
+    O * PROGRAM GOAL
+    O * Perform a CALL passing an array value to a field received as DS
+     V* ==============================================================
+    O * JARIKO ANOMALY
+    O * Before the fix, jariko crashed
+     V* ==============================================================
+     D £C6MIM          DS
+     D  £6I                          20P 6 DIM(99) INZ
+     C                   CALL      'ARRAYDS_1'
```
lanarimarco, line 12: Declare an array of 2 elements, initialize only the first one, and check both elements on return.

Try on AS400 because you are passing a field type (array) different from the type declared in ARRAYDS\_1 entry.

Possibly it's correct, but I want to be sure.

lanarimarco (changes requested):

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/interpreter/values.kt
dom-apuliasoft: See case `MUDRNRAPU00106`/`MUDRNRAPU00107`. In that case we have a DS of 75 to which we assign a string value and want to assign that value to an array value that is 50 (the reminder has to be truncated). In that case `assignableTo` is invoked. 

From a quick glance to `EvalStmt` something different happens when we have an array and when we have a plain value. If you want we can make an in-depth look to find the root cause of the discrepancy but it felt right to always allow string values anyway as they are then correctly evaluated when `coerce` is called.

PR #728, opened 2025-05-16
Bugfix/ls25002246/varying field from ds

dom-apuliasoft:
## Description

Fix an issue that caused `VARYING` fields in `DS` to be longer than expected when their actual length was less than the declared length

Related to:
- LS25002246

## Checklist:


lanarimarco (changes requested):
Add a test also in `ValueTest.kt` where you programmatically create a `DataStructValue` and test whether the `get(data: FieldDefinition)` works as expected.

PR #740, opened 2025-06-17
Bugfix/ls25002734/call indicator external error

dom-apuliasoft:
## Description

Fix an issue that caused `CALL` with indicators to crash with a `null` error.

### Technical notes

The issue derived from the fact that when we had a `CALL` of this kind we popped the last error from the runtime error stack in order not to throw and instead manage it locally.
This worked fine, however if the error was thrown from an external source (for instace a doped program) we tried to pop:
- A wrong error if there was an error prior to the one we caught
- From an empty stack when there was no other error

The latter case is the one that made this issue known.

To fix this a specialized step was added in `CallStmt` that checks if the error we capture is present in the runtime stack before popping and records it in case it was not present (because it means the error originated externally).

Related to:
- LS25002734

## Checklist:


lanarimarco (approved):
Ok.

I have improved the test by checking the indicator value.
See this commit:
https://github.com/smeup/jariko/pull/740/commits/51b91786ff5722fa4d43a38f4d2c5c15ebdcd089

PR #749, opened 2025-07-08
Feature/ls25003049/jariko traces filtering

dom-apuliasoft:
## Description

> Note: This feature does not add/alter any RPGLE functionality.

Add an option to filter out Jariko traces by kind.

### Technical notes

A new optional array of `JarikoTraceKind`s option called `enabledJarikoTraces` was added in the `Options` class. 
It is intended to be a whitelist of all the enabled `JarikoTraceKind`s.

To ensure compatibility with existing systems, by default, when set to `null`, all traces are considered enabled. Use an empty array to disable them all.

> Note: As this behaviour might be counterintuitive the `enabledJarikoTraces` field is private and is not intended to be directly manipulated in the codebase. Use the safe wrappers `enableAllJarikoTraces()`, `getEnabledJarikoTraces()` and `setEnabledJarikoTraces(kinds: List<JarikoTraceKind>)` instead.

Related to:
- LS25003049

## Checklist:

### DIFF
--- a/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/execution/Configuration.kt
+++ b/rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/execution/Configuration.kt
@@ -93,6 +93,7 @@ data class DspfConfig(
  * is true.
  * @param profilingSupport Used to enable/disable scan execution of profiling annotations into rpg sources.
  * This is used to enable the [JarikoCallback.startRpgTrace] and [JarikoCallback.finishRpgTrace] callbacks.
+ * @param enabledJarikoTraces A whitelist of the enabled trace kinds to emit. All enabled if [null].
  * */
 data class Options(
     var muteSupport: Boolean = false,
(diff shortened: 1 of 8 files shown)


lanarimarco (changes requested):
I have some doubts about how to configure this new feature.

Null and empty lists are often seen as the same thing, and more often than not, coders don't read the documentation.

I see this new feature as a callback function rather than a configuration option, for instance:

```kotlin
// return true if kind is enabled
var beforeJarikoTrace: ((kind: JarikoTraceKind) -> Boolean) = {
    true
}
```

This approach allows us to avoid these side effects of your implementation:

1.  Changing the `Options` signature in Java, which does not handle optional fields. This would necessitate changes to new class instantiation and `Options.copy`, and consequently, to the SDK's `JarikoCaller` implementation.
2.  Minor overhead caused by checking `JarikoTraceKind` within a `Collection`. I believe Jariko should minimize overhead whenever possible.
3.  Forcing clients to work solely through `JarikoTraceKind` inclusion.

dom-apuliasoft:
@lanarimarco It seems a nice idea to me. I am going to implement it and if there is no hidden caveat I am replacing the current one with this new one you suggested.

lanarimarco (approved):

PR #765, opened 2025-08-05
Feature/ls25003484/data area support

dom-apuliasoft:
## Description

Add support for data areas through the `IN` and `OUT` statements.

### Technical notes

This PR adds an abstract layer for data areas support. 

This happens through two new customizable callbacks `readDataArea` and `writeDataArea` that are invoked respectively from the `IN` and `OUT` statements. 

Therefore the actual data area read and write implementations are up to the Jariko consumer.

By default:
- `readDataArea` throws an exception and turns on the indicators associated with it.
- `writeDataArea` does nothing.

Related to:
- LS25003484

## Checklist:


rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
```diff
@@ -616,16 +622,17 @@ internal fun ProcedureContext.toAst(
 
     return CompilationUnit(
         fileDefinitions = fileDefinitions,
-        dataDefinitions,
+        dataDefinitions = dataDefinitions,
```
lanarimarco, line 625: Why?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
```diff
@@ -616,16 +622,17 @@ internal fun ProcedureContext.toAst(
...
         main = MainBody(mainStmts, null),
-        subroutines,
+        subroutines = subroutines,
```
lanarimarco, line 627: Why?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
```diff
@@ -616,16 +622,17 @@ internal fun ProcedureContext.toAst(
...
+        dataAreas = mapOf(),
         compileTimeArrays = mutableListOf(),
         directives = mutableListOf(),
         position = toPosition(conf.considerPosition),
         apiDescriptors = null,
         procedures = null,
         procedureName = procedureName,
-        proceduresParamsDataDefinitions
+        proceduresParamsDataDefinitions = proceduresParamsDataDefinitions
```
lanarimarco, line 635: Why?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/utils/misc.kt
```diff
@@ -207,4 +208,11 @@ internal fun getRootCause(t: Throwable): Throwable {
...
     }
     return t
-}
\ No newline at end of file
+}
+
+/**
+ * Exception thrown when something is wrong with data area related operations.
+ */
+data class DataAreaException(override val message: String) : RuntimeException(message) {
```
lanarimarco, line 216: In general, I tend to avoid overriding built-in exceptions, but I might be mistaken in this case. Could you explain the benefits of doing so here?

lanarimarco (changes requested):
Very good work — just a few clarifications needed.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/execution/Configuration.kt
```diff
@@ -353,6 +353,27 @@ data class JarikoCallback(
...
+    /**
+     * It is invoked whenever we want to read a data area.
+     * Returns the value that was read from the data area.
+     *
+     * @param dataAreaName The name of the data area to read.
+     * @param value The current value associated with the data reference.
+     *
+     * @see InStmt
+     */
+    var readDataArea: ((dataAreaName: String, value: String) -> String) = { _, value -> value },
```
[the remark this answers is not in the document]
dom-apuliasoft: That is because I was not very sure we never want to apply different logic depending on the value. Apart from this I am currently using this as trick to keep the value of the data reference unchanged (I pass the current value to the callback and return it from the callback).

Alternatively if it is sure we never want to process the current value, we can remove the `value` parameter and return a `String?` to determine whether or not the value must be changed.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
dom-apuliasoft: The compiler started complaining about mixing positional and named parameters. Very likely I used a wrong order when I added  the `dataAreas` field but given that the issue was real (we were mixing positional and named parameters) I thought it would be better to use a consistent call convention.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
dom-apuliasoft: Same as before

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/utils/misc.kt
dom-apuliasoft: I added this in order to support indicator logic. 
There are cases in which the `IN` statement can return an error but it can be safely managed with the indicator in LO position. 

To catch this behavior we had 3 options:
- Return a magic value from the `readDataArea` callback. I didn't like it because it would cause undocumented multiple checks and validation on the returning value apart from making the interface "dirty".
- Catch any exception (or a specific subset). This works until something goes actually unpredictedly wrong in the implemented custom callback. In this case we just want to turn on an indicator or propagate the error and maybe crash with the error? I think we want the latter.
- Catch a custom exception. This way we can manually throw said exception from the callback and only look for it to turn on indicators, other exceptions fall through.

I might be wrong on the "we want to propagate other errors". If that is the case, please cite it so we can then go back to catching any Exception.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/execution/Configuration.kt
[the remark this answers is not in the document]
lanarimarco: In my opinion, this is not helpful. Regarding the return type being optional, I’m not even sure about that. On the AS400 side, what happens if you try to retrieve a non-existent data area?

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
lanarimarco: See above

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/utils/misc.kt
lanarimarco: Think of the call opcode; the CallStmt implementation is also based on some kind of callback. However, in the call statement—and in general for all statements that activate indicators—we have chosen a simple approach.
Right or wrong, I don’t know, but at least it’s not wrong so far. For this reason, I believe that introducing a new kind of exception is not theoretically incorrect, but it’s not useful and, more importantly, not aligned with the current implementation.
If necessary, we could handle it just as you did with the monitor topic.

lanarimarco (changes requested):
In addition to my comments, I’d like a test of the data area used inside a procedure. Contact Davide because he worked with procedures not too long ago.

rpgJavaInterpreter-core/src/main/kotlin/com/smeup/rpgparser/parsing/parsetreetoast/misc.kt
dom-apuliasoft: I did not change any logic here. We were already passing data definitions to the procedures. The only think that changed is the "style" of params. Before they where a mix of positional and named, now they are all named.

We are passing the same `dataDefinitions`, `subroutines` and `proceduresParamsDataDefinitions` in ProcedureContext CU as we did before.

dom-apuliasoft:
@lanarimarco I am moving this PR status to draft while @davidepalladino-apuliasoft and me investigate the behaviour of missing data areas and data areas inside procedures.

lanarimarco (changes requested):
If I forget to link the data definition with the data area (commented second line):

```rpgle
     D SCAATTDS        DS           460
     C*     *DTAARA       DEFINE    C£C£E00D      SCAATTDS
     C                   EVAL      SCAATTDS='CURRENT'
     C     *LOCK         IN        SCAATTDS
     C     SCAATTDS      DSPLY
```

The error is a `NullPointer,` which makes the diagnosis difficult, see example stacktrace:
```java
Caused by: java.lang.NullPointerException
	at com.smeup.rpgparser.parsing.ast.InStmt.execute(statements.kt:1647)
	at com.smeup.rpgparser.interpreter.InternalInterpreter$execute$internalExecute$1.invoke(internal_interpreter.kt:1206)
	at com.smeup.rpgparser.interpreter.InternalInterpreter$execute$internalExecute$1.invoke(internal_interpreter.kt:1197)
	at com.smeup.rpgparser.interpreter.InternalInterpreter.executeWithMute(internal_interpreter.kt:1269)
	... 63 more
```

Fix and add an error test.

dom-apuliasoft:
@lanarimarco the error message issue should now be sorted out.

lanarimarco (changes requested):
Error tests should follow this pattern:
- Create `ERRORnn.rpgle` where you force an errore
- Implement in `JarikoCallbackTest` the following unit tests:
  - `executeERRORnnCallBackTest`
  - `executeERRORnnSourceLineTest`

PR #779, opened 2025-08-29
Refactor/ls25003611/profiling syntax

enricomasellis-as:
## Description

Refactor profiling annotation syntaxt to be compliant with AS/400 version

Related to:
- ls25003611

## Checklist:


dom-apuliasoft:
@lanarimarco I also want to add a piece of missing context to clarify a design choice that introduces a breaking API change in order to ease the code review process. That is the addition of the `RpgTrace` passed as parameter to the `finishRpgTrace` callback.

The main reason behind this change is the need to support variable captures in annotations (the strange `T(XXX &A(...))` syntax to dump variable data). Those are not bound to the span open but rather to the span close. This forces us not only to know the context in which the span is closed but also to pass data (the variable captures) to the consumer.

If preferred we can only pass the variable captures to `finishRpgTrace` instead of an entire `RpgTrace` to the consumer but passing the entire `RpgTrace` allows the consumer to also have a direct reference on which span has been closed. This might be convenient for the consumer as now we can also specify the span name on span end annotations.

lanarimarco (commented):
I noticed that a test is missing where we demonstrate how to disable tracing for a program.
For example, this could be done by removing the `COP* *TRACE` row or by removing the `CO` prefix. Please add such a test, and also document this behavior in `rpg_profiling.md` under *How to disable program profiling*.

Additionally, I suggest introducing a new callback:

```kotlin
startProgramTrace(program: String): Boolean
```

* It should return `true` if we want to proceed with tracing (default) and `false` otherwise.
* The callback must be invoked on `COP* *TRACE`.

This would allow programmatic control of tracing. For instance, it enables disabling tracing for a specific program without removing the `COP* *TRACE` annotation.

lanarimarco (changes requested):
Some adjustments but in general good work

enricomasellis-as:
@lanarimarco PR is Ready for Review

lanarimarco (approved):

